Issue #12, opened 2025-08-29
What We Do Page


achen2304:
type: marketing site

/locale/what-we-do

combines offerings & impact

PR #31, opened 2025-08-30
feat: early implementation of who-we-are page

achen2304:
# early implementation of who-we-are page
 
- [ ]  You've included unit or integration tests for your change, where applicable.
- [ ]  You've included inline docs for your change, where applicable.

 There's an open issue for the PR that you are making. If you'd like to propose a new feature or change, please open an issue to discuss the change or find an existing issue.

# Description

- Migrated who-we-are page
- updated tailwind configs to match tailwind4

# Notes

- Will most likely need to be updated with new information

#12

### DIFF
--- a/src/app/[locale]/layout.tsx
+++ b/src/app/[locale]/layout.tsx
@@ -97,7 +97,7 @@ export default async function LocaleLayout({
     return (
       <html lang={locale}>
         <body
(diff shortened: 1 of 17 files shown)


dangphung4 (approved):
lgtm

PR #84, opened 2025-09-09
feat: add pages for accesibility, privacy, and terms/conditions

### LINKED ISSUE #22: Accessibility, Privacy, Terms & Disclosure Pages
type: marketing site page

/locale/accessibility
/locale/privacy
/locale/terms 
/locale/disclosures

kurealnum:
## Description

This PR adds the pages for accessibility, privacy, and site terms. All of these pages are accessible, mobile friendly, and support internationalization. See my comments for more information.

Fixes #22


kurealnum:
> One quick thing here, I know Vincent probably talked to you about Japanese translations but adding it here would further complicate the other pages, as they would need to also add the ja translations or else they wouldn't be able to render. my current plan was to add the necessary language support at the end in one big pr once the marketing site is finished though we could try adding something for partially finished translations but imo that goes against having translations in the first place.

I have some unstaged stuff for Japanese translations. Which, the only thing that is actually translated out of all the ja.json files is the Japan privacy policy, which has to be in Japanese. We can do everything in one big PR if that works better though.

achen2304:
> > One quick thing here, I know Vincent probably talked to you about Japanese translations but adding it here would further complicate the other pages, as they would need to also add the ja translations or else they wouldn't be able to render. my current plan was to add the necessary language support at the end in one big pr once the marketing site is finished though we could try adding something for partially finished translations but imo that goes against having translations in the first place.
> 
> I have some unstaged stuff for Japanese translations. Which, the only thing that is actually translated out of all the ja.json files is the Japan privacy policy, which has to be in Japanese. We can do everything in one big PR if that works better though.

see my comment here :p
https://github.com/toddagriscience/Nightcrawler/pull/86#discussion_r2361460025
also @vincenttodd for clarification

kurealnum:
Not really too much to look over here, but still.

achen2304 (approved):
lgtb

PR #196, opened 2025-11-04
build: fix dependabot prs

kurealnum:
Fixes PRs #151 and #158. They will soon be deleted. This updates deps accordingly and changes one field in `tsconfig.json` because NextJS whines about it otherwise (`jsx`).


kurealnum:
going to take a look at the dependabot stuff for this too. Wouldn't it be better to keep everything as up to date as possible though? Within reason, ofc.

Issue #147, opened 2025-11-01
`ProtectedPage` component


kurealnum:
### Add any external media here

This featured adds a `ProtectedPage` component that acts as a wrapper for any protected page (ex. dashboard, some form of account settings, etc.). The component should directly interface and utilize authentication functionality provided in https://github.com/toddagriscience/nightcrawler/issues/144. A user flow is given:
- User accesses a page wrapped by `ProtectedPage`
- If the user is authenticated, they are shown the component the `ProtectedPage` was wrapping (ex. the dashboard)
- If the user is unauthenticated, they are redirected to the `/login` page

PR #197, opened 2025-11-07
feat: authentication logic

### LINKED ISSUE #144: Authentication logic

This issue will implement any necessary authentication logic required to authenticate the user onto the dashboard. This workflow will likely look something like this:

1. User authenticates with Supabase
2. Supabase responds with some form of authentication (JWT, IIRC)
3. User sends a request to backend
4. Backend verifies authentication token (again, JWT IIRC)

Consequently, this issue will have a backend issue linked to it concerning implementing this logic.

kurealnum:
Adds the basics for authentication logic. Login, supabase stuff, environment stuff, idk its 3 in the morning and i still can't sleep but i've been listening to the most fire breakcore ever.

This PR also (temporarily) removes all authentication middleware. Requests will be authenticated to the backend by adding the access token to the `Authorization` headers from https://supabase.com/docs/reference/javascript/auth-getsession. 

Authentication middleware will be _kinda_ reimplemented in #147. Any "blank spots" have a `// TODO ...` left above them and will be handled accordingly.
 
Fixes #144

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -1,8 +1,6 @@
 # Copyright Todd LLC, All rights reserved
 
-# Contributing to Todd Agriscience Website
-
-Thank you for your interest in contributing! This guide will help you get started with the development workflow and project standards.
+Important: Every single file must have "Copyright Todd LLC, All rights reserved." at the top of the file. This may be commented in whatever manner you prefer, but it must be present.
 
 ## 🚀 Quick Start for New Contributors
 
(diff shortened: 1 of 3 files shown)


dangphung4 (dismissed):
Everything is fine so far, test cases and actual implementations should continue in this pr. We shouldn't merge anything that isn't used, like functions or APIs with 0 refs or commented code like that(generally).

I'm keeping this PR up so we can have history for when implementations and further details come.

GitHub is working on stacked diffs so that would be so helpful here

kurealnum:
Ok, did the middleware stuff & (re)-implemented everything accordingly.

src/middleware.ts
artrixdotdev: Convert this to an async function

src/middleware.ts
```diff
@@ -20,8 +20,10 @@ export default function middleware(request: NextRequest) {
   // Check for Global Privacy Control (GPC) signal
   const gpcEnabled = hasGPCEnabled(request);
 
-  // Get authentication status from cookie
-  const isAuthenticated = getAuthStatus(request);
+  let isAuthenticated = false;
```
artrixdotdev: I believe this would always be false since the variable wont be redeclared/updated until after the authentication is finished (io bound operation), and could take a few hundred ms. Because of the fact that `.then` is non blocking, it probably just skips this entirely and will always return false.

vincenttodd (approved):

PR #199, opened 2025-11-10
chore: Migrate completely off of npm in favor of bun

### LINKED ISSUE #29: Move off npm
Just go pnpm or bun

artrixdotdev:
## Checklist

- [X] You've read the [Contributor Guide](https://github.com/toddagriscience/Nightcrawler/blob/main/.github/CONTRIBUTING.md).
- [X] You've included unit or integration tests for your change, where applicable.
- [X] There's an open issue for the PR that you are making. If you'd like to propose a new feature or change, please open an issue to discuss the change or find an existing issue.

## Description
Removes all usages of NPM in favor of [bun](https://bun.com). The following has been completed:

- Migrated all documentation to bun
- All Scripts now use bun
- Devcontainers come preinstalled with bun
- All workflows run with bun

Fixes #29

### DIFF
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -29,7 +29,7 @@ jobs:
         run: bun i
 
       - name: Run unit tests
-        run: bun run test:coverage
+        run: bun run test
 
       - name: Upload test coverage
         uses: codecov/codecov-action@v5


dangphung4:
> Is this accepted by the rest of the team? As in, is this a decision the rest of the team agrees with? (I'm all for bun but removing npm entirely can lead to some edge cases later on)

I know I posted an issue but just wanted to get some eyes on this @vincenttodd

kurealnum:
Just to give my input, I'm fine with moving to bun. We don't have many people on the frontend anyways so the transition wouldn't be that bad.

artrixdotdev:
> My only gripe is that sadly bun doesn't have enough adoption and isn't officially supported with Next(yet). I'm not aware of any other deps we have either

Bun is pretty widely adopted now and has been stable for around 2 years. It's being used at companies like Anthropic and X so I'd say its a pretty safe bet.

As for the Next.js compatibility, this one can get a bit confusing; let me break it down.

Bun offers a few completely separated products, here's a short list:
- A **Bun**dler
- A Compiler
- A **Runtime** (What most people associate bun with)
- A Package manager

^ These are all under the "bun" package

There's probably more I didn't list but this PR **ONLY** takes advantage of the package manager. Next.js still runs on Node.js even when run via `bun dev`. Meaning, when you're actually running the application, there won't be any differences since they're both still using the same runtime. If you *were* to want to use the bun runtime, you would use `bun --bun dev` (confusing I know). Also, Next.js on the bun runtime has been pretty stable as of recent so a future move off of Node.js would be possible, but that's not what this PR does.

TLDR; because of the fact we're only using the bun **package manager**, it will have no affect on the stability of the dependencies or Next.js's ability to run.

vincenttodd:
I'm all for it, sounds great. Thanks for the context too!

PR #235, opened 2025-11-27
feat: login page modal

### LINKED ISSUE #16: Login Page Modal
Edit by Oscar: This component should allow the user to log in with their email and password. We aren't concerned about OTP or OAuth logins at the moment. A user flow is shown:
1. User enters email and password
2. If email and password are validated by the backend, then they are redirected to the `/` page
3. If email and password are incorrect, they are presented with an "Incorrect Email or Password" error.

kurealnum:
Massive PR. Here's what I did:
- Rewrote authentication logic because I was being stupid and was trying to optimistically (using terminology from [here](https://nextjs.org/docs/app/guides/authentication#authorization)) authenticate pages server side. Yes, I know this should've been a different PR. My apologies.
- Changed how intl middleware handles requests. It now simply ignores any un-intl routes such as `/` or `/login`. 
- Implemented a login screen and login functionality via Supabase.
- Implemented a `Protected` component for protected pages. 
- Not wildly important, but I also added a few ShadCN components.

Something to mention: @vincenttodd, authenticated users are currently *not* blocked from going to `/en`. I can change this functionality and check authentication and page load but it'll add another 200-300ms before the page actually renders, which is decreased UX (and possibly poorer SEO?). I can also do some slightly hacky stuff to get around that, but that'll also result in poorer UX for authenticated users.

Closes #16

### DIFF
--- a/src/app/login/page.tsx
+++ b/src/app/login/page.tsx
@@ -1,3 +1,4 @@
+// Copyright Todd Agriscience, Inc. All rights reserved.
 'use client';
 
 import { FadeIn } from '@/components/common';
--- a/src/app/login/page.test.tsx
+++ b/src/app/login/page.test.tsx
@@ -1,3 +1,5 @@
+// Copyright Todd Agriscience, Inc. All rights reserved.
+
 /**
  * @jest-environment jsdom
  */
--- a/src/components/common/utils/submit-button/submit-button.tsx
+++ b/src/components/common/utils/submit-button/submit-button.tsx
@@ -1,3 +1,4 @@
+// Copyright Todd Agriscience, Inc. All rights reserved.
 import { Button } from '@/components/ui';
 import { Spinner } from '@/components/ui/spinner';
 import { useFormStatus } from 'react-dom';
--- a/src/lib/actions/auth.ts
+++ b/src/lib/actions/auth.ts
@@ -1,21 +1,17 @@
+// Copyright Todd Agriscience, Inc. All rights reserved.
 'use server';
 
 import { redirect } from 'next/navigation';
 import { createClient } from '@/lib/supabase/server';
-import { z } from 'zod';
 import logger from '../logger';
 import LoginResponse from '../types/auth';
+import { loginSchema } from '../zod-schemas/auth';
+import { z } from 'zod';
 
-const loginSchema = z.object({
-  email: z.email('Invalid email address'),
-  password: z.string().min(1, 'Password is required'),
-});
+/** This file is STRICTLY for SERVER SIDE authentication. Server-side authentication should be generally preferred over client-side authentication. You can read more about this in `./src/lib/auth.ts` */
 
 /**
- * Logs a user in using Supabase on the server, and handles errors accordingly.
- *
- * DISCLAMER: THIS IS A SERVER ACTION.
- * PLEASE READ THE SERVER ACTION DOCUMENTATION BEFORE CALLING THIS IN RANDOM PLACES.
+ * ONLY USE ON SERVER SIDE! Logs a user in using Supabase on the server, and handles errors accordingly.
  *
  * @param formData The form data containing the email and password
  * @returns {object} - The object described here: https://supabase.com/docs/reference/javascript/auth-signinanonymously
--- a/src/lib/auth.test.ts
+++ b/src/lib/auth.test.ts
@@ -1,3 +1,4 @@
+// Copyright Todd Agriscience, Inc. All rights reserved.
 /**
  * @jest-environment jsdom
  */
--- a/src/lib/auth.ts
+++ b/src/lib/auth.ts
@@ -5,9 +5,13 @@ import LoginResponse from './types/auth';
 import { createClient as createBrowserClient } from './supabase/client';
 import { AuthError } from '@supabase/supabase-js';
 
-/** Auth, assuming credentials are correct, is handled in the following manner:
+/** This file is STRICTLY for CLIENT SIDE AUTH. Unless ABSOLUTELY necessary, prefer server-side auth over client-side authentication for sake of security and leaning into Next.js's standard patterns.
  *
- * 1. User sends a request to Supabase
+ * With that:
+ *
+ * Client side auth + request to API, assuming credentials are correct, is handled in the following manner:
+ *
+ * 1. User attempts to
  * 2. Supabase returns authentication credentials
  * 3. User attempts to access a protected page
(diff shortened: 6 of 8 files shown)


vincenttodd:
That should be fine. I guess the question I would have is, if the user logs in, where is it forwarding after /login (/ or /en)? and if the the user returns to toddagriscience.com, and is already authenticated, what is the / going to be? Is there something like "if authenticated, and /en, then forward to /"--that is only used if /en is used?

Get back whenever!

kurealnum:
@vincenttodd w/ the current setup: When the user logs in (i.e. authenticates successfully), they are redirected to `/`, the current location of the dashboard. If the user navigates to `/en` or `/login` as an authenticated user, they are not redirected to the dashboard, `/`. They can continue to navigate the marketing site, or navigate back to the dashboard (and any other future protected routes). If an authenticated user tries to access the dashboard (or any other protected routes), they are redirected to `/login`. 

Happy to change whatever though.

vincenttodd:
okay, thanks for clarifying. the only marketing pages user would have links to on the dashboard are /privacy, /terms and /disclosures, but if they navigated to those pages, then theoretically nothing would stop them from using the header to navigate elsewhere, which I don't have a problem with. Just if they click the logo to return 'home' (/en) off of, for example, '/en/news' or google search, it forwards to dashboard '/ ' and not '/en' (even if its the dashboard).

So long story short, if it's possible, I'd like to have /en forward to / if user is authenticated. Also, as a safety net possibly, should we 1), if the user has been authenticated, make the "Log In" link on the header (edit: on the marketing site) say "My Account" (with link becoming / from /login), or 2), if the user has been authenticated, should the link '/login' cause them to 'logout', and therefor require them to login again, because I would assume its a bug having the user sitting on a /login page already be authenticated?

kurealnum:
> if the user has been authenticated, make the "Log In" link on the header (edit: on the marketing site) say "My Account" (with link becoming / from /login)

yeah, that's ez. Maybe "Dashboard" instead of "My Account" though? 

> if the user has been authenticated, should the link '/login' cause them to 'logout'

not really a bug. just a quirk, wouldn't really affect anything. If they're logged in though, they can get redirected to the dashboard when accessing `/login`

> if it's possible, I'd like to have /en forward to / if user is authenticated

It is possible, it's just the aspect of user experience again. When accessing a intl route (`/en` for example), either:

Both authenticated and unauthenticated users will see a loading icon, and authenticated users will be redirected to the dashboard.

Neither authenticated or unauthenticated users will see a loading icon (i.e. ~200ms faster load times for all users), but authenticated users will be "yanked" from the page after authentication is done.

I.e, authentication can happen before or after the screen is shown, and there are pros and cons to both. Of course auth (and something of a loading icon) has to happen for all protected routes, but users expect that.

src/components/auth/protected.tsx
```diff
@@ -0,0 +1,56 @@
...
+        <div className="relative h-screen">
+          <FadeIn duration={0.2}>
+            <Spinner className="absolute inset-0 m-auto w-12 h-12" />
+          </FadeIn>
+        </div>
+      );
+    }
+  }
+
+  if (isAuth === true) {
```
artrixdotdev: ?

kurealnum: yes. will fix.

kurealnum:
> the client has to take 2 trips to the server in order to be authenticated opposed to 1

In the way auth is figured in this PR, the client contacts Supabase once to authorize the user (which may be able to be further optimized with `getSession()`, which would only be used for rendering UI components ofc. need to double check that)

> Also, the client side approach produces much more bulky code that's more confusing to understand.

This is true, but opinionated. With proper documentation (and perhaps better code), this can be mitigated for the sake of performance.

A large part of the confusion from everything else stems from the fact that we really don't need NextJS for the way we intend to/are currently building nightcrawler. Authenticated loading of pages can be done client side in a more efficient manner, while still maintaining appropriate levels of security. However I do understand where you're coming from, in that this complicates *everything* a lot more. Let's get someone else's opinion on this.

dangphung4:
Ok, so I see valid points on both sides here.  @kurealnum  is right that we want the application to feel snappy and optimistic. However, @artrixdotdev makes a strong point about code complexity and alignment with the App Router model. In all honesty it doesn't really matter as long as it works, but I'll play devils advocate...

If our goal is the future health of the project, I think we should lean into the framework’s standard patterns:

1. Auth: Let’s move the protection logic to Middleware. While client-side checks are fast, Middleware provides a centralized place for redirect logic (solving the /en -> / requirement) and prevents the “flash of unauthenticated content” issues that become harder to manage as the app grows.

2. Forms: Let’s use the Form component. It handles the heavy lifting for validation state now. We can always look at adopting lighter Field primitives for simpler UI interactions later, but for a critical flow like Login, the robustness of Form is worth it.

Let’s optimize for maintainability now so we don’t have to refactor custom auth hooks later.

kurealnum:
Fair enough. https://ui.shadcn.com/docs/components/form isn't maintained anymore though.

artrixdotdev:
> Fair enough. https://ui.shadcn.com/docs/components/form isn't maintained anymore though.

These might be useful:
https://nextjs.org/docs/app/api-reference/components/form
https://react.dev/reference/react/useActionState

Or
https://tanstack.com/form/latest

kurealnum:
still some stuff to do from here, I'll request a review when I'm done

'
artrixdotdev: ?

src/app/login/page.tsx
```diff
@@ -0,0 +1,123 @@
...
+import { Checkbox } from '@/components/ui/checkbox';
+import { Field, FieldGroup, FieldLabel, FieldSet } from '@/components/ui/field';
+import { Input } from '@/components/ui/input';
+import { Spinner } from '@/components/ui/spinner';
+import { login } from '@/lib/auth';
+import Link from 'next/link';
+import { useRouter } from 'next/navigation';
+import { ChangeEvent, FormEvent, useState } from 'react';
+
+export default function Login() {
```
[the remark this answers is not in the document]
artrixdotdev: If you need help/more explanations on how or why lmk

artrixdotdev (changes requested):
A improvement. I see you moved to middleware, but some of this code could be simplified by using server components and server actions (mainly the ladder).

artrixdotdev (changes requested):
Shouldn't /login be moved to /[locale]/login ?

PR #238, opened 2025-11-30
fix: recreate changes to resolve merge editor issues

### LINKED ISSUE #201: Add `eslint-plugin-license-header`

Adds a plugin for automatically detecting & adding the license header for any source files. This issue should:
- Add [this](https://www.npmjs.com/package/eslint-plugin-license-header) plugin
- Modify the `validate` command to include automatically checking for & adding the license header (or some form of this -- doesn't have to be the validate command, should probably be a separate command)

cinnamon-cicada:
## Checklist

- [ ] You've read the [Contributor Guide](https://github.com/toddagriscience/Nightcrawler/blob/main/.github/CONTRIBUTING.md).
- [ ] There's an open issue for the PR that you are making. If you'd like to propose a new feature or change, please open an issue to discuss the change or find an existing issue.

## Description

I re-pulled and implemented the plugin changes because merge editing wasn't working. We can just delete the other pull request.

Fixes #201

### DIFF
--- a/package.json
+++ b/package.json
@@ -27,7 +27,8 @@
     "prepare": "husky",
     "ci": "bun run type-check && bun run lint && bun run test && bun run build",
     "validate": "bun run format:check && bun run ci",
-    "generate-icons": "tsx scripts/generate-icons.ts"
+    "generate-icons": "tsx scripts/generate-icons.ts",
+    "check-license": "eslint . --rule 'license-header/header: [error, { 'template': '/*\\n * Copyright (c) Todd Agriscience, Inc. All rights reserved.\\n */' }]'"
   },
   "dependencies": {
     "@radix-ui/react-select": "^2.2.6",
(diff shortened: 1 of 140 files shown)


src/app/[locale]/[...rest]/page.tsx
```diff
@@ -1,4 +1,4 @@
-// Copyright Todd Agriscience, Inc. All rights reserved.
+// Copyright (c) Todd Agriscience, Inc. All rights reserved.
```
Comment (kurealnum) [inline src/app/[locale]/[...rest]/page.tsx]: Is that `(c)` preferred now? Totally fine if it is, just wanted to double check!

package.json
```diff
@@ -27,7 +27,8 @@
     "prepare": "husky",
     "ci": "bun run type-check && bun run lint && bun run test && bun run build",
     "validate": "bun run format:check && bun run ci",
-    "generate-icons": "tsx scripts/generate-icons.ts"
+    "generate-icons": "tsx scripts/generate-icons.ts",
+    "check-license": "eslint . --rule 'license-header/header: [error, { 'template': '/*\\n * Copyright (c) Todd Agriscience, Inc. All rights reserved.\\n */' }]'"
```
kurealnum: I saw you have this in `eslint.config.mjs`. The license plugin should run anytime `eslint` is called, so an extra command like this shouldn't be necessary.

cinnamon-cicada: Somewhere, I think I read that we wanted it as a separate check--but if not, I'll remove it

PR #277, opened 2025-12-10
refactor: Forward requests on non protected authenticated users

### LINKED ISSUE #250: Implement 404 routes for authenticated routes

Right now, navigating to a non-existent or incorrect authenticated route while authenticated (ex. `/account/does-not-exist`) redirects the user to `/`. This is not best practice, and should be modified to return a 404. 

This could possibly be done in conjunction with #248.

artrixdotdev:
## Checklist

- [X] You've read the [Contributor Guide](https://github.com/toddagriscience/Nightcrawler/blob/main/.github/CONTRIBUTING.md).
- [X] There's an open issue for the PR that you are making. If you'd like to propose a new feature or change, please open an issue to discuss the change or find an existing issue.

## Description
From what I'm gathering, the issue described in #250 was caused by the auth middleware intercepting requests and auto redirecting users back to `/`. This change uses `NextResponse.next()` to allow the request to "passthrough" as normal so it can be intercepted later on by the router and redirected to the 404 like any other route.

I'm unable to test this at the moment so if someone could that would be great.

Fixes #250


kurealnum:
what other URLs are there other than privacy? lemme fix tests and ill do that also rq

edit: @vincenttodd  and privacy is currently internationalized

kurealnum:
Oh I see what you're saying. Yeah TBH I'll just hardcode that. Easier than trying to do something fancy.

kurealnum (approved):
lgtm!

kurealnum:
note to future readers: I'm only merging this without waiting for Sefu because I want to get this other middleware issue fixed ASAP

PR #283, opened 2025-12-11
feat: integrate sanity add template article page

### LINKED ISSUE #95: Integrate and configure Sanity + template article page
### Problem Statement

A template page for any future articles *on Todd's website* needs to be added. The design of this template page is laid out in Figma (Marketing and Platform), however the general "rubric" looks something like this:
- SEO (and consequently, performance, accessibility, and general best practices) are **extremely** important. OpenGraph metadata, structured data, and anything else that could potentially help with SEO should be added in this PR. [This blog post](https://www.artisan.co/blog/ai-sdr), according to Vincent, is highly ranked in terms of SEO, so maybe consider using it as "inspiration"
- Each article should be listed under `/news/{article-name}
- A CMS should be integrated in this PR. https://www.sanity.io/ should preferably be used as it was requested by Vincent, but I am open to suggestions.

kurealnum:
Closes #95 

This is an extremely large PR. About half of it comes from Sanity's studio (~15 files) which is installed in the root of the project at `./studio`. Some notes:
- Some logic and requests could potentially be optimized in the future. Specifically, using eager/lazy loading in different situations. For sake of shipping though, this is fine.
- OG and Twitter metadata are not currently handled for each specific news article.
- Styling is minimal and can be updated in the future
- Sanity provides an optional (and tedious) way to provide types for your documents. This is something that, unless someone else thinks otherwise, I'd like to do in the future. See [here](https://www.buildwithmatija.com/blog/how-to-generate-typescript-types-for-your-sanity-v3-schema) for a short tutorial. The one potentially major upside is the ability to use `useDocument`, which *sounds* like the proper way to retrieve documents. However, Sanity's [tutorial](https://www.sanity.io/docs/next-js-quickstart/diplaying-content-in-next-js#k632ccaa580f3) also gives another way to fetch documents, which is the way Sanity is currently set up.

One other big thing: Jest is becoming more and more of a pain, especially now that we've integrated Sanity. While I was the one who originally wanted to stick with Jest (oops!), I would like to move to Vitest as I'd think it'd solve the majority of problems I've been having ([see](https://stackoverflow.com/questions/61781271/jest-wont-transform-the-module-syntaxerror-cannot-use-import-statement-outsi), this has been an issue previously, among other things). I'll test this locally at somepoint and see if it would work. If this is alright with everyone, the big question is when to move to Vitest. The easiest (and least "proper") way would be to move everything over on this PR.

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -4,7 +4,50 @@ This is the general contributing guide for the marketing site and platform. Plea
 
 ## Contents
 
-1. **adding this might help once its done**
+- [Basic Workflow](#basic-workflow)
+  - [Notes](#notes)
+- [When Things Break](#when-things-break)
+  - [The 4 C's](#the-4-cs)
+- [Development Workflow](#development-workflow)
+- [First-Time Setup](#first-time-setup)
+  - [Prerequisites](#prerequisites)
+  - [Getting Started](#getting-started)
+- [Standards / General Information](#standardsgeneral-information)
+  - [Testing](#testing)
+  - [Git Hooks Validation](#git-hooks-validation)
+  - [Formatting / Styling](#formattingstyling)
+  - [Running Tests](#running-tests)
+  - [Test Patterns](#test-patterns)
+- [Code Standards](#code-standards)
+  - [TypeScript](#typescript)
+  - [Styling](#styling)
+  - [File Organization](#file-organization)
+- [Logging Guidelines](#logging-guidelines)
+  - [Using the Logger Utility](#using-the-logger-utility)
+  - [Logging Best Practices](#logging-best-practices)
+  - [When to Log](#when-to-log)
+- [Internationalization (i18n)](#internationalization-i18n)
+  - [Adding New Text](#adding-new-text)
+  - [Translation Guidelines](#translation-guidelines)
+- [Accessibility Requirements](#accessibility-requirements)
+  - [Standards](#standards)
+  - [Testing](#testing-1)
+- [Component Development](#component-development)
+  - [For Shared Components](#for-shared-components)
+  - [Component Structure](#component-structure)
+  - [Add Tests](#add-tests)
+  - [Add Storybook Story](#add-storybook-story)
+- [Development Tools](#development-tools)
+  - [Essential Commands](#essential-commands)
+  - [Git Hooks](#git-hooks)
+- [Common Pitfalls](#common-pitfalls)
+  - [Avoid These Mistakes](#avoid-these-mistakes)
+  - [Performance Considerations](#performance-considerations)
+- [Pull Request Guidelines](#pull-request-guidelines)
+  - [PR Checklist](#pr-checklist)
+  - [PR Description](#pr-description)
+- [License](#license)
+- [Code of Conduct](#code-of-conduct)
 
 ## Basic Workflow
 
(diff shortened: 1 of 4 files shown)


src/app/[locale]/news/page.tsx
```diff
@@ -4,14 +4,32 @@
...
-import { useNews } from '@/lib/hooks/news';
+import sanityQuery from '@/lib/sanity/query';
+import { SanityDocument } from 'next-sanity';
+import { useEffect, useState } from 'react';
 
 /**
  * Highlighted news & general news
  * @returns {JSX.Element} - The news page
  */
 export default function News() {
```
Comment (artrixdotdev) [inline src/app/[locale]/news/page.tsx]: Why isn't this a server component? It would remove the need of the useState and useEffect. Also, If you add a loading.tsx it would probably result in faster load times. If this cant be a server component, use the new [`use`](https://react.dev/reference/react/use) function provided by react to clean this up

src/components/common/news/featured-news-carousel.tsx
```diff
@@ -10,20 +13,37 @@ import NewsCardProps from '../news-card/types/news-card';
...
           key={article.title}
           isDark={false}
-          image={{ url: article.image.url, alt: article.image.alt }}
+          image={
+            article.thumbnail
+              ? {
+                  url: urlFor(article.thumbnail)?.url(),
+                  alt: article.thumbnail.alt,
+                }
+              : { url: ArticlePlaceholder.src, alt: '' }
```
artrixdotdev: Just use /article-placeholder.webp, no need for the import shenanigans

src/components/common/share-article/share-article.tsx
```diff
@@ -0,0 +1,108 @@
...
+  TooltipTrigger,
+} from '@/components/ui/tooltip';
+import { useState } from 'react';
+
+/** Primarily used for `/en/news/[slug]`. Provides a few buttons that automatically share articles to social platforms. Facebook is the one exception: due to needing to make an account, among other things, with Facebook's developer platform, clicking on the Facebook icon only copies the link of the article to your clipboard and displays a small "Copied!" dialog.
+ *
+ * @param {string} title - The title of the article.
+ * @returns {JSX.Element} - The buttons, wrapped with a fragment
+ * */
+export default function ShareArticleButtons({ title }: { title: string }) {
```
artrixdotdev: Split these up into their own components so the code is easier to understand

artrixdotdev (changes requested):
Looks fine so far.

Although, since studio is its npm project (it has its own package.json) it should be added to the root package.json to create a monorepo. Normally it isn't a good idea to have multiple package.jsons lying around without any kind of connection. Would link them with a basic monorepo config with bun 

https://bun.com/docs/guides/install/workspaces

Should be a 1 line fix

src/components/common/share-article/share-article.tsx
kurealnum: ig, gonna keep them in the same file though

src/components/common/news/featured-news-carousel.tsx
kurealnum: typing thing here. if the name or location of `article-placeholder.webp` changes in the future, this catches it at compile time. small issue, i know, but worth considering.

artrixdotdev: Move them outside the public directory then and into a local images folder in the src directory to make the import less sloppy looking.

kurealnum: that adds unnecessary complexity to solve a nitpick

vincenttodd (approved):

PR #344, opened 2026-01-06
refactor: rework contact form

### LINKED ISSUE #270: Implement Inbound Onboarding (Frontend)

The current contact form (see [here](https://toddagriscience.com/en/contact)) is sufficient for general use cases, but we'd like to rework it to better fit the new onboarding process. You can get a general gist of how this will look on the Figma on the `Marketing and Platform` tab, but here's a quick use case:

1. User navigates to `/{locale}/contact`
2. User inputs first name, last name, farm name, work email, and phone number
4. User navigates to the next page of the form. If they enter a work email, allow them to skip the website input.
5. User enters the farm's website. If they don't have a farm website, they aren't a fit.
6. User navigates to the next page of the form
7. User answers yes/no to the "organic or biodynamic" question
8. If the user "is a fit" (see notes), save their information and display something along the lines of a success page. If the user isn't a fit, say "Todd might not be the right fit" and move on.

This should be completed with a ShadCN carousel.

The new contact form should contain a honeypot.

Some notes:
- Non-work emails should be explicitly filtered out
- Phone numbers should be validated and formatted appropriately
- If the user does not have a work email and does not have a website, they are not a fit.
- When the user officially signs up, cookie-based tracking (see PostHog) should be enabled. This is legal. Don't ask me why.

Please save a copy of the old contact form under `templates/` in the documentation repository for future reference (https://github.com/toddagriscience/colossus/tree/main).

kurealnum:
## Description

Closes #270. This is not complete (hence the draft PR). Any backend or authentication logic is still not implemented. At the moment, only the UI is implemented, which features:
- A form in a carousel for the first name, last name, farm name, email, phone number (carousel page 1), the farm website if a business email is not given (carousel page 2), and the type of farm (carousel page 3)
- A helper function for determining a "business" email
- A few packages to support this page, namely ShadCN Carousel (Embla) and polyfills for testing

## Checklist


vincenttodd:
might benefit to add something that if someone enters a email, phone or website, its a actual email address with a `@` and `.`, 10 digit phone number with the `+1` only btw, or website and not just plan text as I just did. I'm sure there could be weird domains, but maybe just requiring a `.` for the domain, not sure?

Also super petty, but for a time being, could you change the organic/biodynamic question to: "Do you grow using organic, biodynamic or regenerative principles?"

vincenttodd:
Another thing I remembered, can we add analytics like this? It's posthog, but not sure how it's done. Might be for another issue tbh.

516557539-f2022603-d385-4f2f-9c1f-c13556a87778

vincenttodd:
thats fine. We don't need it yet anyways

PR #374, opened 2026-01-14
feat: implement wildcard support for protected routes in auth middleware

### LINKED ISSUE #248: Update `middleware/auth.ts` logic to allow for wildcards

Right now, to add a protected route, you have to manually add the exact route to `protectedUrls` in `middleware/auth.ts` which is inconvenient and error-prone. The logic in this file should be updated to support wildcards so that the following use cases function appropriately:

```ts
// Blocks unauthenticated users from `/accounts/home`, /accounts/somewhere`, and so on.
const protectedUrls = ['/', '/accounts/*'];
```

CristianGramada:
## Description
Updated `middleware/auth.ts` to support wildcard patterns in protected routes, eliminating the need to manually add each route.

Fixes #248 

## Checklist

### DIFF
--- a/frontend/README.md
+++ b/frontend/README.md
@@ -159,6 +159,15 @@ If you'd like to:
 
 We utilize multiple different UI libraries as well as Tailwind v4. Global styles are contained in `./src/app/globals.css`
 
+#### Background Colors
+
+The site uses two separate background colors:
+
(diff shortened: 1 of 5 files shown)


kurealnum:
Once that documentation change is made, lgtm, will get it approved.

PR #473, opened 2026-02-02
feat: Form saving logic for `/accept`

### LINKED ISSUE #431: Add form saving logic for `/accept`
### Problem Statement

This issue should add an action for the `/accept` page that updates any pre-existing information and saves any new information, as well as updates the user's password via Supabase. This action should return an error if:
- The password is invalid
- Any other data is invalid

### Context

This issue acts as the backend logic for inviting a user

### Constraints

This issue should not:
- Let the user change their role (Admin or Viewer)

### Technical Details

Take a look at other actions in the nightcrawler repo and see if they could be abstracted into a single action used by multiple other actions. I'm pretty sure the password setting logic is used more than once across the repository

When the user's password is changed *successfully* and the user's data is saved *successfully*, the user should be logged in.

If the user's email is not verified (see `isVerified()`) return an error immediately. If the user's password is saved successfully, set `verified` to true.

### User Story

_No response_

### Use Case

_No response_

artrixdotdev:
## Description

Fixes #431 

## Checklist

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -8,10 +8,13 @@
         "@typescript-eslint/eslint-plugin": "^8.54.0",
         "@typescript-eslint/parser": "^8.54.0",
         "@vitest/eslint-plugin": "^1.6.6",
-        "esbuild": "^0.27.2",
+        "esbuild": "^0.25.0",
         "eslint-plugin-drizzle": "^0.2.3",
         "eslint-plugin-no-secrets": "^2.2.1",
-        "storybook": "^10.2.1",
+        "lodash": "^4.17.23",
+        "lodash-es": "^4.17.23",
+        "storybook": "^10.2.3",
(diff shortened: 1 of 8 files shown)


frontend/src/app/(authenticated)/accept/actions.ts
```diff
@@ -0,0 +1,109 @@
...
+
+  if (!verified) {
+    logger.warn('Attempted to accept invitation for unverified user');
+    return {
+      error:
+        'User email is not verified. Please verify your email before accepting the invitation.',
+    };
+  }
+
+  const rawData = {
```
kurealnum: This needs to have all the relevant fields in the user model, namely `didOwnAndControlParcel` and `didManageAndControl`. Use a type created from the table.

PR #496, opened 2026-02-05
feat: tabs

### LINKED ISSUE #448: Implement tab system on platform
### Problem Statement

This issue should implement tabs for navigating through management zones and the homepage. You can refer to the Figma for more information on how a tab should look. Refer to [this FigJam](https://www.figma.com/board/vv6ObZY25RoFWFD2xovpVH/Todd-Whiteboard?node-id=0-1&p=f&t=sndZ2iaId7ToEqCs-0) for exact use case information (under Tab Management Flow, bottom left).

The tab system is somewhat self explanatory with the above information. It's essentially a CRUD API that interfaces with ShadCN's tabs. Note that each tab will need to have a bit of information associated with to pass to the underlying component (which, as described later in this issue, will not be implemented in this PR). 

Please stack PRs with each subissue.

### Context

Tabs are described in further detail in the [platform design docs](https://github.com/toddagriscience/colossus/blob/main/design/platform.md#tabs):

> After logging into the platform, the user lands on a new tab, or pre-existing first open tab.
> 
> The largest part of the page is the Function Area. It is used for viewing widgets, data, news, products or support.
> 
> Above the function area is Tab Bar where the client can select their management zone or other static tabs. Selecting the management zones allows us to set the ambient data. Clients can open multiple tabs.
> 
> The "kinds" of tabs that a user can open are:
> 
>  - Management zones
>  - Farm management (AKA the entire farm)
>  - Homepage (not something that the user can "open", but it exists in a tab)

### Constraints

This issue should NOT:
- Implement any widgets or pull any information from the farm

### Technical Details

- ShadCN's tabs should be utilized here
- Tabs should be saved so the tabs are the same when the user gets back
- Each time the user creates or deletes a new tab, database entries will need to be updated
- Tabs should be able to be renamed by clicking on the tab itself and modifying the default name
- A maximum of X tabs should be allowed at a time (TBD)

### User Story

As a user, I want to be able to click "+" to open a new tab, so I can view multiple management zones on a single webpage. 

### Use Case

- User opens platform
- User is shown homepage
- User navigates to pre-saved tab
- User clicks "+" to open a new tab
- Dropdown appears
- User clicks on "Corn Field"
- Dropdown closes
- New tab for Corn Field is opened

kurealnum:
## Description

Fixes #448. Didn't end up going with the "X" showing only on hover - it looked kinda awkward. Plus, having the "X" on every tab mimics the browser (AKA an environment that users are familiar with).

## Checklist

### DIFF
--- a/frontend/src/app/(authenticated)/page.tsx
+++ b/frontend/src/app/(authenticated)/page.tsx
@@ -1,9 +1,6 @@
 // Copyright © Todd Agriscience, Inc. All rights reserved.
 
-import { 
-  accountAgreementAcceptance, 
-  managementZone 
-} from '@/lib/db/schema';
+import { accountAgreementAcceptance, managementZone } from '@/lib/db/schema';
 import { db } from '@/lib/db/schema/connection';
 import { tab } from '@/lib/db/schema/tab';
 import { getAuthenticatedInfo } from '@/lib/utils/get-authenticated-user-farm-id';


kurealnum:
> Whatever happened, it's a slight regression from the pervious version. Also not sure if hover was suppose to be on the last preview.
> 
> Also, I feel like having the x visible on all the tabs seems bold/aggressive. Don't mind it visible on the tab the user is on, but idk. @clam365 or @kyyamaa do you guys have an opinion? Maybe for accessibility. Safari hides the x unless you hover on the tab.
> Screenshot 2026-02-05 at 2 24 48 PM
> 
> Not sure if you can decipher any info from this, but I want the same padding for us.
> Screenshot 2026-02-05 at 7 19 12 PM

Slight regression in what way? And wdym same padding? 

I'm still not a big fan of only showing the X on hovering. That might be the standard for Safari, but it isn't for Firefox or Chrome. I do want to try making it a bit smaller. I also need to hop back in and modify some accessibility stuff.

vincenttodd:
> A homepage is necessary IMO. Gives the client a sense of direction and control. In the future, we can give a brief overview of the farm, management zones, etc. Could go the route of making it not deletable though? 

^ answered.

> Why make the application mandatory before getting access to the platform? Having the platform visible gives a nice preview of everything. If there's ever a certain page or two that we don't want clients accessing before being accepted, we can just add a check for that.

No okay, thats true. But we would need to disable the search feature, because that would still work without providing farm profile info, as far as I know. That was my concern, to give away data without them paying or us knowing who they are (i.e. competitors). I really like Palantir Foundry in this aspect, it's free for people, but in reality it's just a way to hook people on wanting to use their paid services _through_ that platform.

(edit: I probably would like clients to search products though.)

> I understand what you're saying about Robinhood, but for what it's worth, neither of us have any idea how a client would feel about this (he-said she-said kinda thing?). As far as the search bar goes, that'll probably end up being pressed to the right side. I know that most of our current clients have < 10 management zones, but I worry about clients that potentially have more.

Thats a valid point, but I mean, anything that looks to out-of-scale for a professional "platform", is just want I want to avoid. We don't need a big header area, and defiantly not a large wordmark, that can stay at the same scale as the tabs. Robinhood emulates a more professional vibe, just compared to ours, simply bc our header area is quite large. But it's just a minor cosmetic tweak at the end, don't sweat it rn.

PR #510, opened 2026-02-09
refactor: getCurrentUser / getAuthenticatedInfo error handling

### LINKED ISSUE #435: Refactor `getCurrentUser()`
### Problem Statement

The `getCurrentUser()` function returns the current user from the database if authenticated, and returns an object with an error in it otherwise. This method of handling errors isn't the best however, and could be refactored. For instance, this is what the current error handling looks like:

```ts
  const currentUser = await getAuthenticatedInfo();

  if ('error' in currentUser) {
    // handle error
  }
```

This could be handled with either an actual `Error` or `null` (`Error` would be preferred). Other ideas are welcome since this function will be used constantly across the repository.

When this change is made, please update all uses of this function accordingly.

### Context

This function will be widely used across the repository, so making it as simple and easy to use is important. This is purely a DX change.

### Constraints

_No response_

### Technical Details

_No response_

### User Story

_No response_

### Use Case

_No response_

CristianGramada:
## Description

Fixes #435 

## Checklist

### DIFF
--- a/frontend/src/app/(authenticated)/apply/components/colleagues/action.ts
+++ b/frontend/src/app/(authenticated)/apply/components/colleagues/action.ts
@@ -34,11 +34,7 @@ export async function resendVerificationEmail(
   email: string
 ): Promise<ActionResponse> {
   try {
-    const result = await getAuthenticatedInfo();
-
-    if ('error' in result) {
-      return result;
-    }
+    await getAuthenticatedInfo();
 
     await resendEmailInvite(email);
 


frontend/src/app/(authenticated)/page.tsx
```diff
@@ -29,35 +27,35 @@ export const metadata: Metadata = {
  * @returns {React.ReactNode} - The dashboard page component
  */
 export default async function DashboardPage() {
-  const currentUser = await getAuthenticatedInfo();
+  let currentUser: Awaited<ReturnType<typeof getAuthenticatedInfo>>;
```
kurealnum: Same thing here

frontend/src/lib/utils/get-authenticated-user-farm-id.ts
```diff
@@ -33,18 +45,20 @@ export async function getAuthenticatedInfo(): Promise<
...
-    return currentUser;
+    return { ...currentUser, farmId: currentUser.farmId };
   } catch (error) {
-    if (error instanceof Error) {
-      return { error: error.message };
+    if (error instanceof Error && isKnownAuthError(error)) {
+      throw error;
     }
-    return { error: 'Error with querying for the current user.' };
+    throw new Error('Error with querying for the current user.', {
```
kurealnum: I see what you're trying to do, but just throw an error and log any extra information with `logger`.

PR #529, opened 2026-02-11
fix: remove auth navbar on forgot-password

### LINKED ISSUE #459: Fix authenticated navbar showing on `/forgot-password`
### Problem Statement

The authenticated navbar is showing on `/forgot-password`, which causes user confusion. The public navbar should be shown instead since the user isn't authenticated if they're accessing the `/forgot-password` route.

Also, while you're here, please center the reset password form vertically.

### Context

_No response_

### Constraints

_No response_

### Technical Details

_No response_

### User Story

_No response_

### Use Case

_No response_

cinnamon-cicada:
## Description

Fixes #459 

## Checklist


cinnamon-cicada:
Oh, got it! I thought we wanted the non-authenticated header AND did not want it under a locale. Will change it.
________________________________

kurealnum:
Oh that's my bad. I just looked at the issue again and you're right. But yeah atp just change it so there's no header, like I mentioned above. Easier for all parties involved.

PR #582, opened 2026-02-19
feat(platform): add internal contact page (#544)

### LINKED ISSUE #544: Add internal contact page
### Problem Statement

This should add a *basic* internal contact page. It should provide an email (service.us@toddagriscience.com) and phone number (+1 (888) 279-1283).

I don't have time to make a mockup at the moment, but it should look something like this:

<horizontally and vertically centered div>
Contact us at any time.

email here

phone number here 
</>

Text should be left aligned.

### Context

In the future this will be much more robust, but this'll do for now.

### Constraints

_No response_

### Technical Details

_No response_

### User Story

_No response_

### Use Case

_No response_

inbanr:
## Description

Fixes #544 

## Checklist

### DIFF
--- a/frontend/src/app/(authenticated)/contact/layout.tsx
+++ b/frontend/src/app/(authenticated)/contact/layout.tsx
@@ -1,0 +1,15 @@
+// Copyright © Todd Agriscience, Inc. All rights reserved.
+
+import { Metadata } from 'next';
+
+export const metadata: Metadata = {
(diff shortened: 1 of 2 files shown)


kurealnum (commented):
Before this gets merged, make sure the page is centered, both horizontally and vertically with respect to the space that the navbar takes up. Also add a title via metadata (place this in a `layout.tsx` file in the same folder).

PR #631, opened 2026-02-24
feat: authenticated contact page

### LINKED ISSUE #566: Implement client-facing `Contact` page
### Problem Statement

At the moment, `/contact` just shows a form. It should first show a page, as shown on `/contact` in the Figma. The mockup on Figma will need to be finalized before this is implemented.

### Context

This is an internationalized page, public to all users.

### Constraints

_No response_

### Technical Details

- The client facing contact page (think of it as the landing page for `/contact` - AKA not the form itself) should be on the same page as the form itself. You might want to utilize query params or lazy loading for this.

### User Story

As a potential client, I want to view information about Todd & getting in contact, so that I can feel confident in my choice to work with Todd
As a potential client, I want to land on `/contact` and see information about Todd, then click a button to navigate to a form where I can provide more information

### Use Case

_No response_

vincenttodd:
## Description

This was just something I threw together rq. We can delete it if it's trash. Just thought it might be a little help.

Fixes #566 

## Checklist


akeight:
> > yeah this is fine, need to make font a bit smaller though. make it look prettier basically. can take a look if you don't wanna/can't
> 
> @kurealnum You can, I think this would be a good Extern project tho. Chat does horrible with frontend. With all these pages I've made today, I'm just putting up the code, the frontend needs major renovations.

I mean....it works for those with vision impairments...

PR #632, opened 2026-02-24
fix: auto frontend install before no verify

### LINKED ISSUE #627: [Bug]: Needing to push code with no-verify
### Bug Description

I've been doing this for a while now (probably not a good thing?), but since I have the extra credits, I asked why, here's the list:

These are the repo-wide failures from the pre-push validate hook (format:check && ci, where ci runs type-check, lint, test, build). The block came from type-check:

- validator.ts missing generated route modules

Missing dependency type/module errors (these are obviously new):

- widgets-grid.tsx -> react-grid-layout
- actions.ts -> react-grid-layout
- widget.ts -> react-grid-layout
- layout.tsx -> schema-dts
- macro-radar-widget.tsx -> recharts
- metric-range-tooltip.tsx -> recharts, recharts/types/component/DefaultTooltipContent
- mineral-level-widget.tsx -> recharts
- chart.tsx -> recharts
- dropdown-menu.tsx -> @radix-ui/react-dropdown-menu
- tabs.tsx -> @radix-ui/react-tabs

I can provide more info if needed.

### Steps to Reproduce

bun run validate

### Expected Behavior

not needing to push no-verify?

### Screenshots/Videos

_No response_

### Browser

Safari

### Device Type

Desktop

### Language/Locale

None

### Additional Context

Both Codex and Cursor have the same issue.

vincenttodd:
## Description

This fixes the no-verify issue for AGENTS.md (the context file used by OpenAI). I confirmed it cannot/does not use .corsorrules or CLAUDE.md. Those files were also updated.

Fixes #627

## Checklist


frontend/.cursorrules
```diff
@@ -155,6 +155,13 @@ When making changes, prioritize:
 
 ### Development Workflow
 
+#### Commit/Push Reliability Workflow
+
+- Before any commit or push flow that uses `--no-verify`, automatically run: `cd frontend && rm -rf .next && bun i`.
```
kurealnum: didn't realize you were gonna put this here, just say something like this:

Before any commit or push flow that uses `--no-verify`, automatically run `bun validate` in the `frontend` folder.

vincenttodd: Chat told me to do that. I thought bun validate just checks, not installs? no-verify also happens on cursor, and the only fix in the past was running `cd frontend && rm -rf .next && bun I`. I'm happy to update, just want to double check I understand first.

vincenttodd: I don't mean to keep going, but every time I push a commit, I have the issue of needing to push no-verify, it never fails. I've made it a habit to just run it every time. I guess this must be my computers problem, not the terminal? Theres no way for me to push code without the no-verify unless I install.

vincenttodd:
If this is good, I want to merge this and update AGENTS with more stuff.

PR #728, opened 2026-03-04
feat: header image loading effect

### LINKED ISSUE #458: Add blur/loading effect to Header Image Component
### Problem Statement

A new header image component was added for articles in #417. This image is extremely large quality-wise and needs to have a loading animation added so that it doesn't snap onto the screen. 

### Context

_No response_

### Constraints

This issue should not change the size or shape of the header image component.

### Technical Details

_No response_

### User Story

_No response_

### Use Case

_No response_

chuhuihan:
## Description

Fixes #458 

## Checklist

- [X] You've included unit or integration tests for your change, where applicable.
- [X] Any components that you've modified are accessible.
- [X] You've used [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/) where appropriate


vincenttodd:
@kurealnum pretty good, like the idea. I think the animation might need shortened, its almost a full second before seeing the image, it feels kinda like its buffering

PR #787, opened 2026-03-19
feat: implement seed products

### LINKED ISSUE #785: Implement seed products
### Problem Statement

This issue implements seed products. These seed products should appear both on the search bar as well as `/product/{seed-slug}`.

#### Searching for seeds

Seeds should be able to be searched for inline with IMPs. A new seed table will need to be created with an embeddings vector of length 3072 in order to support this search of 2 different data types. Cosine similarity should still be used.

#### Viewing seeds

The following information should be presented for each `/product/{seed-slug/`:

- Seed name
- General information about the seed
- Supply/stock
- Cost per unit
- Picture (will require S3 integration via Cloudflare's R2)

And the following buttons should be added:

- Order now
- Way of contacting an advisor 

#### Shopping cart

All items that are "purchased" will first be added to a shopping cart. The cart will be labeled to the user as "Order" at all times. The status of the Order can be maintained via the local storage API. The Order can be displayed under `/order`. When there are items in the shopping cart, the navbar item `Order` should be shown. 

#### Checkout screen

When the user is ready to check out, they'll click on the shopping cart icon, which will ask them to confirm if they're done. If they are, they'll input their payment information (make sure to specify that we never store payment information), and pay us. Then, an order will be placed. Leave this function blank for now. 

### Context

This provides the ability to directly & efficiently purchase seeds without having to ask an advisor about stock. It should provide a generic interface that is simple and easy to work with.

### Constraints

_No response_

### Technical Details

- The database seeding script will need to be updated
- New migrations will need to be made
- A script should be added in the database package that allows developers to programmatically add new seed variants to the development database

### User Story

As a user, I want to be able to effectively and efficiently browse seeds so that I can purchase seeds from Todd

### Use Case

_No response_

kurealnum:
## Description

Fixes #785. Still need to do code review on all of this.

## Checklist


apps/site/src/app/(authenticated)/(misc)/order/components/order-client.tsx
```diff
@@ -0,0 +1,394 @@
...
+      return;
+    }
+
+    if (processedSessionIdRef.current === sessionId) {
+      return;
+    }
+
+    processedSessionIdRef.current = sessionId;
+    const returnedSessionId = sessionId;
+    let isActive = true;
```
kurealnum: document

apps/site/src/app/(authenticated)/(misc)/order/components/order-client.tsx
```diff
@@ -0,0 +1,394 @@
...
+              <div>
+                <h1 className="text-3xl font-semibold text-foreground">
+                  Order
+                </h1>
+                <p className="mt-2 text-foreground/70">
+                  Review your selected seed products before checkout.
+                </p>
+              </div>
+
+              {order.items.map((item) => (
```
kurealnum: component this

apps/site/src/app/(authenticated)/(misc)/order/page.tsx
```diff
@@ -0,0 +1,24 @@
...
+export default async function OrderPage() {
+  const currentUser = await getAuthenticatedInfo();
+
+  if (!currentUser.approved) {
+    notFound();
+  }
+
+  return (
+    <OrderClient
+      stripePublishableKey={
```
kurealnum: not necessary to pass through

apps/site/src/components/ui/dialog.tsx
kurealnum: revert this, not needed (maybe, double check me)

apps/site/next.config.ts
```diff
@@ -4,6 +4,13 @@ import type { NextConfig } from 'next';
 import createNextIntlPlugin from 'next-intl/plugin';
 
 const withNextIntl = createNextIntlPlugin();
+const seedImageHostnames = (process.env.NEXT_PUBLIC_SEED_IMAGE_HOSTNAMES ?? '')
```
kurealnum: change this so the environment variable is a single base URL, not just a list of hostnames. then remove this function entirely.

packages/db/src/schema/seed-order-checkout.ts
```diff
@@ -0,0 +1,51 @@
...
+/** Purchased quantity for a single seed product within a processed checkout. */
+export interface SeedOrderCheckoutItem {
+  /** Seed product id purchased during checkout. */
+  seedProductId: number;
+  /** Purchased quantity for the matching seed product. */
+  quantity: number;
+}
+
+/** Processed Stripe checkout session for a seed order. */
+export const seedOrderCheckout = pgTable('seed_order_checkout', {
```
kurealnum: is this necessary? I think not. stripe API should return all relevant information

kurealnum:
what what do you mean??? i did my best

PR #837, opened 2026-04-11
fix: replace auth "Thinking..." loading with shared spinner

CristianGramada:
## Description

Fixes #827 

## Checklist


vincenttodd (approved):
looks good! if you need to update packages, try to keep them in a separate pr so we can roll back easy if theres a problem. But it's fine this time.

PR #845, opened 2026-04-27
fix: /apply form to use ach payments + tests

### LINKED ISSUE #841: Refactor application payment form and backend
### Problem Statement

currently users are required to enter credit card details and subscribe to continue the application. Access to the platform should be free once the client application is approved unless they decide to work with us as a paid client. 

No one can access the platform with the current config without subscribing.

### Context

We need to collect the client's bank information during the application for ACH subscription payments at a later date. Stripe allows us to collect and store information with profile information within their system. The information is accessible to clients in under their `/account` page. 

Based on my limited knowledge, it looks like system emails for payments is handle in Stripe's system. I will work on getting a sub domain set up for that.

I'd prefer the "elements" UI components, but if it's easier to use the pre-built option, we can do that.

### Constraints

- This shouldn't change any of the other forms fields
- This shouldn't charge clients up front

### Technical Details

- https://docs.stripe.com/payments/ach-direct-debit
- https://docs.stripe.com/payments/ach-direct-debit/set-up-payment?payment-ui=elements

### User Story

_No response_

### Use Case

- User should enter ACH details and continue their application without becoming subscribed or charged.

yousefalwahami:
## Description

Fixes #841 

Replaces the old paid-subscription step in the `/apply` flow with an ACH bank-information collection step. Farms now connect a bank via Stripe Financial Connections (instant verification) so they can be billed later without being charged *today*, and `/account` shows their bank on file even on the free tier.

Summary/highlights:
- New `Bank Information` apply tab using Stripe Elements (`SetupIntent` + `us_bank_account`), with custom `@font-face` loaded into the iframe so it matches the site font.
- Server actions `createAchSetupIntent` / `recordAchSetupComplete` plus a `setup_intent.succeeded` webhook handler that pins the new bank as the customer's default payment method.
- `/account` now resolves a payment method from the customer (default -> most recent) instead of only from an active subscription.
- Fixed the **post-apply redirect** so users who've already submitted aren't bounced back to `/apply`.
- Test coverage: 33 new tests across server actions, Stripe utilities, the webhook helper, and the component.

## Checklist


vincenttodd (commented):
@akeight or @dgonz10152 if either of you want to check this pr, it looks good from what I can tell, but always like a second eye on big prs

CristianGramada (dismissed):
LGTM!

vincenttodd (approved):

PR #847, opened 2026-04-29
feat: add canonical /index/[slug] CMS articles with legacy /news redirects

### LINKED ISSUE #290: Implement Sanity types

The way that Sanity is currently integrated in `nightcrawler` doesn't allow for accurate document or GROQ query types (i.e, `myArticle.title` is not type checked). Types will need to be manually generated via [Sanity TypeGen](https://www.sanity.io/docs/apis-and-sdks/sanity-typegen) for both the schemas and GROQ queries that we currently utilize.

A note on GROQ queries: currently, we're manually querying Sanity with the `sanityQuery` function. With types added however, we should be able to utilize `useDocument()`/`useDocuments()` ([see](https://www.sanity.io/learn/course/build-content-apps-with-sanity-app-sdk/use-document)) instead of explicitly writing queries. Please research this *before* starting work on this issue.

Finally, if `useDocument()`/`useDocuments()`  is utilized and there is still room for performance improvements, a new low priority issue should be created regarding Sanity performance improvements.

### LINKED ISSUE #842: Refactor CMS backend for `/news/` articles
### Problem Statement

Currently there is only one route for all news/press, stories, product releases, research, etc; `/news/[slug]`

### Context

The goal is to set up the route and article template to handle several types of news. The new route will be `/index/[slug]`. These articles will be accessible under `/research`, `/stories`, `/news` and any other page that relates to news, stories, products or research.

### Constraints

`/index/[slug]` should only use one template.

### Technical Details

There is external news being managed on the same system as in-house written articles. When the visitor clicks on an external news link, it's temporary routed through `/news/[external-article-title]` to the final destination. These external articles are not index in the dynamic sitemap. When updating the template route, factor in the update of these external articles routes and potentially excluding these from sitemap inclusion.

The current `/news/[article]` page template will be used for this. Some modification will need to happen to page structure to accommodate the added information and page images. This can be handled/cleaned up by frontend.

Categorization may be something that needs set up with this for organizing articles on different parent pages like `/research` or `/news`. 

### User Story

A user should have a consistent experience across different news types. Accessing a product release should feel similar to access a client story.

### Use Case

_No response_

yousefalwahami:
## Description

Fixes #842 

This change implements the single shared article surface at `/{locale}/index/[slug]` so press, stories, product notes, research, and similar content types can share one template while listings under `/news`, `/research`, and future hubs can deep-link consistently.

Summary of what shipped:
- Sanity CMS: Schema and types extended for categorization and metadata (contentType, collections, related fields as designed) so the frontend can distinguish article kinds and assemble listings.`
- Routing: Canonical article route under `index/[slug]`; `/news/[slug]` kept as redirects toward the canonical URL so bookmarks and inbound links stay valid while URLs converge on `/index/` for the template.
- External articles: Readers hit a Todd URL such as `/news/[slug]` that immediately sends them to the real article elsewhere. The sitemap only includes real on-site articles (e.g. under `/index/…`), not these shortcut/redirect URLs.
- Frontend: Listing components, URL helpers, and server actions/query paths updated so cards and links resolve to `/…/index/{slug}`; article shell/metadata adjusted for the unified template where needed.

## Out of Scope
**Note** that this does not close #290, #842 stays on routes, CMS fields, and frontend wiring; Sanity TypeGen (typed schemas + GROQ, and eventual `useDocument` / `useDocuments` instead of hand-written `sanityQuery`) is still tracked on #290.

## Checklist

### DIFF
--- a/apps/sanity/schemaTypes/news.ts
+++ b/apps/sanity/schemaTypes/news.ts
@@ -4,12 +4,73 @@ import {DocumentTextIcon} from '@sanity/icons'
 import {format, parseISO} from 'date-fns'
 import {defineField, defineType} from 'sanity'
 
+/** Shared article model — `_type` remains `news` for migration parity. */
 export default defineType({
   name: 'news',
-  title: 'News Article',
+  title: 'Article',
   icon: DocumentTextIcon,
   type: 'document',
   fields: [
+    defineField({
+      name: 'contentType',
+      title: 'Content type',
+      type: 'string',
+      initialValue: 'news',
+      options: {
+        list: [
+          {title: 'News', value: 'news'},
+          {title: 'Research', value: 'research'},
+          {title: 'Story', value: 'story'},
+          {title: 'Product release', value: 'product-release'},
+          {title: 'Press', value: 'press'},
+        ],
+        layout: 'radio',
+      },
+      description:
+        'Primary category for routing and parent pages (newsroom, research, stories, etc.). Defaults to news for older posts.',
+    }),
+    defineField({
+      name: 'collections',
+      title: 'Additional collections',
+      type: 'array',
+      of: [{type: 'string'}],
+      options: {
+        list: [
+          {title: 'News', value: 'news'},
+          {title: 'Research', value: 'research'},
+          {title: 'Story', value: 'story'},
+          {title: 'Product release', value: 'product-release'},
+          {title: 'Press', value: 'press'},
+        ],
+        layout: 'tags',
+      },
+      description:
+        'Optional extra groupings when an article should appear on more than one parent page.',
+    }),
+    defineField({
+      name: 'canonicalParent',
+      title: 'Canonical parent (optional)',
+      type: 'string',
+      options: {
+        list: [
+          {title: 'News', value: 'news'},
+          {title: 'Research', value: 'research'},
+          {title: 'Story', value: 'story'},
+          {title: 'Product release', value: 'product-release'},
+          {title: 'Press', value: 'press'},
+        ],
+      },
+      description:
+        'Preferred parent for breadcrumbs or analytics when an article spans multiple collections.',
+    }),
+    defineField({
+      name: 'excludeFromSitemap',
+      title: 'Exclude from sitemap',
+      type: 'boolean',
+      initialValue: false,
+      description:
+        'Internal articles only: when enabled, no XML sitemap entry is generated. External articles are always excluded.',
+    }),
     defineField({
       name: 'title',
       title: 'Title',
@@ -182,7 +243,7 @@ export default defineType({
       title: 'Off Site Link',
       type: 'url',
       description:
-        'If a URL is present in this field, no news page will exist for this article. Specifically, navigating to a URL such as `/en/news/{your-slug}` will redirect to this URL.',
+        'When set, navigating to `/[locale]/news/{slug}`, `/[locale]/index/{slug}`, or older legacy links redirects to this external URL.',
     }),
     defineField({
       name: 'isFeatured',
--- a/apps/site/src/app/(unauthenticated)/[locale]/(marketing)/components/news-highlights/news-highlights.tsx
+++ b/apps/site/src/app/(unauthenticated)/[locale]/(marketing)/components/news-highlights/news-highlights.tsx
@@ -5,13 +5,28 @@
 import { FadeIn, NewsCard } from '@/components/common';
 import { Spinner } from '@/components/ui/spinner';
 import { useTheme } from '@/context/theme/ThemeContext';
-import sanityQuery from '@/lib/sanity/query';
+import { loadArticlesForHighlights } from '@/lib/sanity/article-actions';
+import type { SanityArticle } from '@/lib/sanity/article-types';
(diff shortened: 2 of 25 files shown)


apps/sanity/schemaTypes/news.ts
```diff
@@ -4,12 +4,73 @@ import {DocumentTextIcon} from '@sanity/icons'
...
+      title: 'Content type',
+      type: 'string',
+      initialValue: 'news',
+      options: {
+        list: [
+          {title: 'News', value: 'news'},
+          {title: 'Research', value: 'research'},
+          {title: 'Story', value: 'story'},
+          {title: 'Product release', value: 'product-release'},
+          {title: 'Press', value: 'press'},
```
vincenttodd, line 25: on these defined fields, would we be coming back to update for the actual types:
```
Publications
Conclusion
Milestone
Release
Company
Research
Product
Global Affairs
Stories
```

apps/site/src/lib/sanity/article-types.ts
```diff
@@ -0,0 +1,70 @@
...
+
+import type { PortableTextBlock } from '@portabletext/types';
+
+/** Primary article classifications used across the site and Sanity. */
+export const ARTICLE_CONTENT_TYPES = [
+  'news',
+  'research',
+  'story',
+  'product-release',
+  'press',
```
vincenttodd, line 11: > on these defined fields, would we be coming back to update for the actual types:
> 
> ```
> Publications
> Conclusion
> Milestone
> Release
> Company
> Research
> Product
> Global Affairs
> Stories
> ```

same question as above

apps/site/src/app/(unauthenticated)/[locale]/(marketing)/components/news-highlights/news-highlights.tsx
```diff
@@ -5,21 +5,36 @@
 import { FadeIn, NewsCard } from '@/components/common';
 import { Spinner } from '@/components/ui/spinner';
 import { useTheme } from '@/context/theme/ThemeContext';
-import sanityQuery from '@/lib/sanity/query';
+import { loadArticlesForHighlights } from '@/lib/sanity/article-actions';
+import type { SanityArticle } from '@/lib/sanity/article-types';
+import { getArticleCardHref } from '@/lib/sanity/article-urls';
 import { urlFor } from '@/lib/sanity/utils';
```
Comment (vincenttodd) [inline apps/site/src/app/(unauthenticated)/[locale]/(marketing)/components/news-highlights/news-highlights.tsx:11]: just out of curiosity, are these new imports? Oscar use to get mad when I would make unnecessary imports, not that these are, but just genuinely curious what these are

vincenttodd (commented):
Another biggie! 2 questions but otherwise lgtm. Also works in prod.

apps/site/src/app/(unauthenticated)/[locale]/(marketing)/components/news-highlights/news-highlights.tsx
Comment (yousefalwahami) [reply apps/site/src/app/(unauthenticated)/[locale]/(marketing)/components/news-highlights/news-highlights.tsx:11]: good question! 
`loadArticlesForHighlights` fetches the list, `SanityArticle` types the payload, `getArticleCardHref` builds the `/…/index/…` link, and `urlFor` is for the Sanity image refs, we're referencing them all in the file.

apps/sanity/schemaTypes/news.ts
yousefalwahami: So yes we could come back and extend this. `news.ts` defines what editors pick for content type, collections, and canonical parent when using Sanity. This file is pretty much defining labels, fields, validations, dropdown options, etc.

I could add these now if you'd like, but would just need more clarification on what we need and dont need (e.g. in the list Stories vs currently using Story, Release vs Product Release). Another thing, are these all "content types" or also additional fields on the article schema? Like say company, publications, milestone, conclusion sound like tags/collections.


`article-types.ts` is the same value strings, so on the frontend fetching and types will line up, should keep those two files in sync.

vincenttodd: k, we can come back since I don't really have the content for those extra types just yet.

apps/site/src/components/common/news/latest-news-table.tsx
```diff
@@ -47,11 +60,7 @@ export function LatestNewsTable({ items }: LatestNewsTableProps) {
           <div>{item.title}</div>
 
           <Link
-            href={
-              item.offSiteUrl
-                ? item.offSiteUrl
-                : windowHref + '/' + item.slug.current
-            }
+            href={getArticleCardHref(item)}
```
cristiangramada, line 63: Found one routing issue in LatestNewsTable: internal article links can lose the locale prefix. 
On /es/news, an internal row may route to /index/{slug} instead of /es/index/{slug}.
Suggested fix: use locale-aware Link from @/i18n/config for internal URLs and keep <a> for external offSiteUrl links.

CristianGramada (changes requested):

apps/site/src/components/common/news/latest-news-table.tsx
yousefalwahami: Good catch! working as expected now

CristianGramada (approved):
LGTM!

vincenttodd (approved):

PR #871, opened 2026-05-06
feat: errorhanding + home button

john-yang-11:
## Description
created error handling when there were rendering issues for unauth. Also created the /home instead of logout for auth when there is an error
Fixes #{850}

## Checklist


yousefalwahami:
Resolve merge conflicts please

PR #870, opened 2026-05-05
feat: careers CMS listings, split sitemaps, and robots updates

yousefalwahami:
## Description

Fixes #844 

We now treat `careers` like `news`, same Sanity and the same `/[locale]/index/[slug]` article UI, while listing careers content on `/[locale]/careers` and splitting XML sitemaps so crawler traffic isn't one giant mixed URL set.

#### CMS & types
- Add `careers` as `contentType`, `collections`, and `canonicalParent` on the Article (`news`) schema in Sanity, and extend `ArticleCollection` / `ArticleContentType` on the site.

#### Careers Route
- Refactor `/careers` so the Todd University block lives in `CareersExternship`, with an optional server-rendered block for **Featured + Latest** career articles (same carousel/table primitives as `/news`).
- Add `/[locale]/careers/[slug]`: internal posts 308 to `/[locale]/index/[slug]`; `offSiteUrl` still redirects externally (parity with `/news/[slug]`).

#### Sitemaps & robots.txt
- Use generateSitemaps so Next serves `/sitemap/main.xml` (static routes + non-career `/index/[slug]`) and `/sitemap/careers.xml` (career-classified `/index/[slug]` only).
- Replace the single `Sitemap:` line in `robots.txt` with `sitemap/main.xml` and `sitemap/careers.xml` (`/sitemap.xml` isn’t used as XML in this setup).
- Sanity fetch helpers: `getMainSitemapArticles` / `getCareersSitemapArticles` (+ `isCareerArticle`).

## Checklist

- [x] You've included unit or integration tests for your change, where applicable.
- [x] You've included inline docs for your change, where applicable.
- [x] Any components that you've modified are accessible.
- [x] You've used [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/) where appropriate

PR #872, opened 2026-05-07
feat: careers job template

### LINKED ISSUE #843: Create a jobs template with CMS system
### Problem Statement

We are reliant on LinkedIn for all our job posting needs. They've recently degraded their service and can no longer rely on them. 

### Context

The goal is to set up a page template and CMS handling system for job postings. It will act in the same manner as the `/news/` (or depending on when this is worked on: `/index`), but be listed under `/careers/[job-slug]`

### Constraints

`/careers/[job-slug]` should only use one template.

### Technical Details

Temporary we'll use the article template that already exists. 

These jobs will be listed under the upcoming page `/careers/index`. Temporarily these just won't be accessible on the site.

### User Story

Visitor should be able to explore all our open jobs, and apply. 

### Use Case

_No response_

yousefalwahami:
## Description

Builds off #870 , merge first

Fixes #843 + separate `Careers` from `Article` in Sanity Studio

### Sanity Studio

- Added a `career` document type (Careers in the CMS) for job postings, sharing the same body/SEO/portable-text shape as `news` via `article-shared-fields.ts`.
- Article (`news`) no longer shows Careers in Content type / collections / canonical parent so jobs live under Careers only; legacy `news` docs that already use careers tagging still resolve on the site until migrated. (hasn't been used, but good practice for prod)

### Marketing site

- Broadened Sanity fetches so listings, detail slug resolution, and sitemap slices treat `news` and `career` as one logical “article” shape (projection coerces routing fields where needed).
- `isCareerArticle` treats `_type === 'career'` as careers content and still supports legacy careers-tagged `news` rows.
- `getArticleBySlug` queries both types with deterministic ordering when slug collisions occur.

### Routes

- `/careers/index` permanently redirects to `/careers`.
- `/careers/[slug]` continues to `notFound` for non-career documents and otherwise matches existing redirect behavior.

## Checklist


vincenttodd (approved):
looks good

PR #891, opened 2026-05-20
feat: new home page

akeight:
## Description

Fixes #829 

## Checklist


akeight:
@vincenttodd The only difference from what we had is that the buttons now have a hover color applied to them after the merge from main. Do you want them like that or adjusted?

vincenttodd (approved):

Issue #890, opened 2026-05-20
Refactor application flow


vincenttodd:
### Problem Statement

currently we're asking for operational information before giving them permission to access the platform which is just redundant. Its also too many steps for onboarding.

### Context

We're going to move all farm operational questions to pre-account creation or through email. The new onboarding flow is: `email magic link` > set password page > invite there users > add payment details, including billing and separate farm/shipping address (optional/if needed) > accept terms > full access to the platform.

We will no longer need the 'application under review' banner or supporting systems related to the review period.

### Technical Details

if possibly, it might be better to comment out/screenshot the old application form to keep it for records rn, there are several questions that will potentially need to be incorporated into the request form or elsewhere on the platform. 

On the new onboarding flow, make sure the client can't access the platform without filling it out, including to access the account page. It might be best to just remove the header, since we'll be direct touch with clients pre-onboarding, and if they have questions they will email us directly. 

Ref figmas for visual and layout

PR #893, opened 2026-05-27
Feat/request form + Refactor application flow

### LINKED ISSUE #888: Refactor signup/request form
### Problem Statement

From feedback from multiple users, here are the current standings:

- Theres no way to resend the verification email
- Once the Yes/No is selected, make the page automatically go to the next page
- Change the “join us” to “continue”
- On the address form on the application, typing the first letter make the cursor jump to the next form field
- Business website is not required to be collected on the application, not to mention its required but no *
- Save and next button on application does not work

most of these won't be relevant with the update to the signup flow.

### Context

The new client flow to signup will be:

`/index/[article-slug]` > "Request access" button > `/forms/[platform-access]` > submit form to client services + review > if approved, email signup link > set password page > add other users to account > add payment details > accept terms and conditions > platform is setup

### Constraints

_No response_

### Technical Details

_No response_

### User Story

_No response_

### Use Case

_No response_

yousefalwahami:
## Description

Fixes #888 #890 

Adds CMS platform-access forms and an internal applications dashboard, then wires approved applicants through a single magic-link onboarding path (no duplicate signup email, signed in after password).

### Forms & data
- Public `/forms/[slug]` with Sanity-driven fields, retention consent, and smart email/name/phone extraction from arbitrary CMS field names
- Create table to track applications (`platform_access_application` table); migration `0008`
- `platform_access_application` table + signup tracking (`invite_sent_at`, `signed_up_at`, `farm_id`); migration `0009`

### Internal dashboard
- Review approve/reject, form-slug filter, resend magic-link invite, signup status + fallback link copy

### Approved applicant onboarding
- Approve/resend -> Supabase magic link -> `/incoming` -> `/signup` (session required)
- `setPassword` + `signIn` -> redirect `/apply` (mirrors colleague accept flow)
- Token validation, phone normalization (+1 stripped in URLs), gate UI if session missing

### Also in branch
- multi-placement article CTAs buttons, careers/article schema cleanup

### Tests I've done
- [x] Submit a platform-access form locally; confirm row in internal `/applications`
- [x] Approve -> magic link email -> `/incoming` -> password -> lands on `/apply` logged in (sends one email only)
- [x] Resend invite after cleanup; stale /signup URL shows gate without session
- [x] `bun run db:local:migrate` applies 0008/0009 on fresh DB

## Checklist


longlong9889:
Local verification + findings (pulled #893 down and ran the full flow)

  Request form → DB (the Monday-critical path): 
  Status: works ✅

  - Submitted the live Sanity form (iris-access) locally → got "Request Received" → confirmed the row persisted in the platform_access_application table (status pending).
  - Type-check + lint pass. Cold production build (--force, 0 cache) compiles clean, exit 0, all new routes present.
  - This is what Monday needs according to Vincent. Clients can submit requests.

 Post-submission onboarding: 3 real bugs (all after the request form; none block Monday)

  The magic-link onboarding (approve → email → set password → …) doesn't complete end-to-end. Traced it to a chain of three code bugs:

  1. auth/confirm/route.ts — assigned a query-bearing next (/incoming?...&token=...) straight to .pathname, which escapes the ? into the path → 404 right after login, token lost. Root cause found; small fix (parse next into path + query) applied locally.
  2. Onboarding "confirm info" → /signup — navigates to a non-localized /signup (no /en/), which 404s on client-side navigation. Found, not yet fixed.
  3. /signup missing from the allUserRoutesIntl allowlist — once the magic link authenticates the user, the auth proxy redirects them off /signup (a localized route) to /, which errors since the account isn't provisioned yet. This is the wall that stops completion. Found; one-line fix, but it touches the auth middleware so leaving it for sign-off.

cristiangramada (dismissed):
Strong work on the core pipeline. The CMS forms, internal dashboard, magic-link approval flow, token validation, and prefill extraction all look solid. #888's pre-account flow (article -> request access -> review -> signup) is in good shape.

Several #890 items still need attention before we can call this issue done:

1. #890 - ApplicationReviewBanner was not removed

Issue #890 says we no longer need the "application under review" banner. It's still imported and rendered in the authenticated layout.

2. #890 - No platform gate until onboarding is complete

After approved signup, users land on /apply, but there's no global guard stopping them from going to /account, /search, /order, etc. while onboarding is incomplete. #890 expects clients not to access the platform (including account) until onboarding is done.

3. Data consistency bug in approved-applicant signup

In signup/actions.ts, completeApprovedApplicantSignup writes to Supabase first (setPassword, updateUser, signIn) and only then calls persistSignupRecords. If the DB step fails, the user can end up authenticated in Supabase with no matching user/farm row. Either reorder so DB writes happen first where possible, or add rollback/cleanup on failure.

4. Missing tests for the approved-applicant path

actions.test.ts / page.test.tsx cover the cold signup path, but not the magic-link flow (applicationId + token, session gate, completeApprovedApplicantSignup, redirect to /apply). That's the most critical new behavior. Please add coverage.


#890 scope still open

Slim down post-signup /apply. Issue describes magic link -> password -> invite users -> payment -> terms; the full 5-tab flow (General Business, Colleagues, Farm, Bank, Terms) is unchanged.
Archive old application form. Issue suggested commenting out / preserving the old form for records; /apply is still live as-is.
Remove header during pre-onboarding. /apply still uses the authenticated header via (misc)/layout.tsx.


Nits

redirect('/contact') on the signup page drops locale. Signup lives under [locale]; redirect should preserve it (e.g. /en/contact).
Hardcoded "Select an option" in form-field-renderer.tsx. Should go through i18n like the rest of the form.
Internal detail page still shows the signup link after signedUpAt is set. Confusing since the token is rejected at that point; hide or disable once signup is complete.
incoming/page.tsx is 'use client' though it only reads search params. Could be a server component.

vincenttodd:
good spots @CristianGramada. since post account approval flow needs reworked from a structural standpoint too, it this is def not done. I'll need to merge this anyways for the pre account flow today, but I'll keep the issues open and circle back to this later in the week.

vincenttodd (approved):

PR #898, opened 2026-05-28
feat: replace tab navigation with management zone sidebar

### LINKED ISSUE #894: Refactor platform management zone tab menu
### Problem Statement

Current platform allows client to open/close management zone “tabs” like a browser.

### Context

Clients should have a more convenient way to view and navigate to each management zone page. Every management zone on the farm (that has a soil test performed) get its own page. The information for each zone page is from separate db rows (I think).

Basically we’re going to take the management zone tab selector component (clicking on the + icon shows the list of management zones for a given farm) and convert that to a left side of window menu list. Clients should have the ability to re arrange the list to their preference. Default sort can be like by time added, oldest on top and newest on the bottom.

The left tab menu should not interfere with the header (TODD wordmark and links). Visually it can resemble like the ChatGPT recent chat list or codex app file list. Final frontend tweaks will happen once the other components are added. 

When the client first loads the homepage, it should default open the first management zone page, or even ideally the last management zone the client was on (not important for this issue). The other pages should not fully load until they are clicked on.

The tab component can be commented out, but not removed. I have another use case for that. 

### Constraints

Each management zone page currently is lazy loaded and only fully loaded when clients navigate to a given tab. That can remain the same. 

Clients also have the option to rename a management zone name by clicking directly on the name. That feature should also remain. 

### Technical Details

_No response_

### User Story

Clients should always see their full list of management zones

### Use Case

_No response_

longlong9889:
## Description

 Fixes #894
                                                                                                                                                                                    
  - Replaced browser-style tab navigation with a persistent left sidebar listing all management zones
  - Zones are displayed with green dot indicators and keyboard shortcut badges                                                                                                      
  - Supports inline renaming for admins by clicking the active zone name
  - Supports drag-to-reorder via @dnd-kit, persisted to localStorage
  - Lazy loading of zone content preserved, only the selected zone loads
  - Old tab component code commented out, not removed (has future use case according to Vincent)

  ## Checklist

  - [x] You've used [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/) where appropriate)


longlong9889:
hello, i requested access to Vercel to see deployment issue, my account is longnguyen7803-7082

yousefalwahami (changes requested):
LGTM, just 1 issue + something I'd look over with @vincenttodd :

I'm getting a hydration mismatch issue, it's coming from `zone-sidebar.tsx`: `@dnd-kit` runs during SSR, so sortable rows get mismatched `aria-describedby` IDs (`DndDescribedBy-3` vs `DndDescribedBy-0`). Also reading `localStorage` in the initial `useState` can reorder zones before hydration. Fix: static list on first render, then mount dnd-kit + saved order in `useEffect.`

The url `/?zone=#` is using the database `management_zone.id` (from `management-zone.ts`. This is fine I guess but this means the badges on the sidebar (like id 1, 2, 3, ..) aren't actually the same url (so we could be on the first management zone but the url could be `/?zone=201`. Not breaking anything or causing problems, but something to consider for design.

yousefalwahami (approved):
lgtm

vincenttodd (approved):

PR #915, opened 2026-06-04
feat: replace /search page with slide-in search results panel

### LINKED ISSUE #897: Create search module
### Problem Statement

Currently when clients use the search bar and click enter, they are directed to `/search` (I think) and shown the results.

### Context

The goal is when a client enters a search, a module slides from the right side of the page showing the results. We want to remove the `/search` page completely. The module should be able to open and close with a collapse button without losing the search results unless the client searches for something new. 

### Constraints

_No response_

### Technical Details

_No response_

### User Story

_No response_

### Use Case

_No response_

longlong9889:
## Description

Fixes #897

 ## What changed

  - **New slide-in search panel** (`components/search-panel/`)
    - Opens automatically when a search is submitted
    - Collapses to a slim rail via a collapse button **without losing the current results**
    - Reopens on demand, keeping the previous results until a new search is run
    - Shows loading skeletons, an error state, and a "contact a Todd advisor" message when there are no matches
  - **Rewired the header search input** (`search-nav-form.tsx`) to open the panel (`runSearch`) instead of navigating to `/search`
  - **Removed the `/search` page entirely** — deleted the route; its result-card, skeleton, and types were moved into `search panel/`
  - **Wired the panel into the dashboard** (`(widgets)/page.tsx`) via a `SearchPanelProvider` so state (query/results/open) is shared between the header input and the panel

  ## Behavior

  - Typing in the header search shows the live dropdown preview (unchanged).
  - Pressing Enter / clicking the search icon opens the slide-in panel with the full results.
  - The collapse button shrinks the panel to a rail but keeps the results; only a new search replaces them.

  ## Testing notes

  - The panel UI and the header dropdown render regardless of environment.
  - **Live results require** the database to have knowledge data seeded **with embeddings** and `OPENAI_EMBEDDINGS_KEY` set. Without the key, the panel correctly shows the "Search failed" state -> that's environmental, not a bug


vincenttodd (approved):

C0derTang (changes requested):
```
{
      "file": "apps/site/src/app/(authenticated)/(misc)/product/[slug]/page.tsx",
      "line": 40,
      "summary": "Hardcoded href=\"/search\" points to deleted page",
      "failure_scenario": "User clicks 'Back to search' from a product detail page → 404. The /search route was deleted and no redirect was added."
    },
    {
      "file": "apps/site/src/app/(authenticated)/(misc)/imp/[slug]/page.tsx",
      "line": 53,
      "summary": "Hardcoded href=\"/search\" points to deleted page",
      "failure_scenario": "User clicks 'Back to search' from an IMP detail page → 404. The /search route was deleted and no redirect was added."
    }
```

might cause 404 bc the search page got deleted, lowk dont know how to test it on my side tho

PR #917, opened 2026-06-05
feat:  implement research index listing page

### LINKED ISSUE #916: Implement /research/index page
### Problem Statement

The marketing site needs a research/index page. Implement it from the Figma design.

### Context

_No response_

### Constraints

_No response_

### Technical Details

_No response_

### User Story

_No response_

### Use Case

_No response_

dshal20:
## Description

Fixes #916 

## Checklist


C0derTang:
### I did a thing

The page is finished and working, it just has no articles to show yet (apart from my sample one), so it displays **"No entries yet."**

The page only pulls in articles marked as **Research** in Sanity. Right now none of the 12 articles in Sanity are marked that way (they're tagged as press, product release, story, or news instead) idk if u wanted a new "research" type of article specifically or u wanted all the old articles to show up aswell

**To make articles show up:** in Sanity Studio, set an article's **Content Type** to **Research** (or add **research** to its Collections).

The category buttons at the top are smart too: a button only shows if there are articles of that type. So once Research articles exist, the right buttons appear on their own.

If you want me to show other stuff as well (news stories) its a p simple change

C0derTang:
### What changed from the original commits (in `f2299ef`)

Building on @dshal20's original page, this commit fixes two things from review. In plain terms:

**1. Uses real data, not fake rows**
- Before: the page showed 10 made-up articles hardcoded in the file.
- After: it pulls the real articles from Sanity. The fake list is gone.

**2. Rows are clickable now**
- Before: rows looked clickable but did nothing.
- After: each row links to its article (internal page, or opens the external link in a new tab). Same helper the news list uses.

**3. Filter tabs match real categories**
- Before: tabs were made-up labels (Publication / Conclusion / Milestone / Release) that don't exist in Sanity.
- After: tabs come from the real Sanity types (Research / Story / Product Release / Press), and a tab only shows if at least one article of that type exists.

**4. Works in both languages**
- Before: all text was English only, so the Spanish page showed English (against our rules).
- After: text moved into translation files (en + es). `/es/research/index` now shows Spanish.

**5. Tests rewritten**
- Before: tests checked the fake titles.
- After: tests use mock data and cover the heading, tabs, links, pagination, filtering, and the empty state. 7/7 pass.

**Not done (on purpose):** the Filter / Sort / grid-list buttons are still just labels with no action; no Storybook story (we don't have page stories in this repo).

One thing to note: dates now show as `23 April 2026` instead of `Apr 23, 2026`.

C0derTang:
### Update (commit `d719fdf`)

Turned the research page into a reusable template and added a few things:

**1. One shared page layout**
The research list layout is now its own component (`ArticleIndex`), so other pages can reuse it instead of copy-pasting the markup.

**2. Each category has its own page**
Categories are now real pages with real links: `/research/index/research`, `/research/index/story`, etc. Clicking a tab takes you to that page (instead of changing a hidden setting in the URL). A category that doesn't exist shows a 404, and each one gets its own browser-tab title.

**3. The News page now matches**
`/news` was rebuilt to use the same layout as the research page so they look the same. It doesn't show category tabs (the `/news/[slug]` article links already use that part of the URL), and the old carousel was removed.

**4. Text is shared between both pages**
The button/label text (Filter, Sort, View more, etc.) moved into one shared translation file (`articleIndex`), so research and news both use it — in English and Spanish.

**5. Security fix (XSS)**
Links coming from Sanity were trusted as-is. A bad link like `javascript:...` could run code when someone clicked it. Added a small checker (`toSafeHref`) that only allows normal links (http/https, mailto/tel, internal paths) and drops anything dangerous. Applied it everywhere a Sanity link is shown — list rows, article buttons, and in-text links — plus a matching check in Sanity Studio.

**6. Tests**
Added/updated tests for all of the above, including one proving a `javascript:` link never renders as a clickable link. All passing.

akeight:
I haven't been able to dig into the files changed, but it looks and works as expected. Sanity is all set up for it as it should. Great job!!

vincenttodd (approved):
somebody with Sanity experience needs to refactor some of this older stuff, but great job on it!

Issue #895, opened 2026-05-28
[Bug]: PII Exposed in URL Parameters, GDPR/CCPA Violation


vincenttodd:
### Bug Description

Full name, email, and phone number are passed as GET query parameters instead of the POST
request body. This data is permanently written to Cloudflare access logs, Vercel server logs,
browser history, and the HTTP Referer header. PostHog analytics captures these full URLs on
every submission, including failed ones.

### Steps to Reproduce

/en/signup?first_name=...&last_name=...&email=...&phone=...

### Expected Behavior

Move all form fields to the POST request body immediately. Audit and request log deletion from
Cloudflare, Vercel, and PostHog for URLs containing PII. Review PostHog data retention policy.

### Browser

Chrome

### Device Type

Desktop

### Language/Locale

None

### Additional Context

GDPR/CCPA violation. PII is unrecoverably logged across multiple systems on every signup
attempt. Confirmed live, full name, email, and phone visible in Cloudflare and Vercel logs. Any
team member or vendor with log access can enumerate all applicants.

PR #924, opened 2026-06-11
Feat/platform onboarding gate

yousefalwahami:
## Description

Fixes platform-access onboarding end-to-end: CMS submissions now land in `form_submissions`, hydrate farm tables on signup, and flow through a token-only `/signup` link without PII in URLs (addresses [#895](https://github.com/toddagriscience/nightcrawler/issues/895)).

**Database & hydration**
- Renames `platform_access_applications` -> `form_submissions` (migration `0011`) and adds `farm.advisor_profile_notes`.
- Maps `iris-access` answer keys to Postgres tables on signup via `hydrate-farm-from-form-submission` + `iris-access-field-map` allowlists.
- Signup tokens and context live in `form-submission-signup` queries (`resolveSignupContext`, `completeFormSubmissionSignup`).

**CMS forms (iris-access)**
- Adds `checkboxGroup` field type (certifications, alternate farming, near contamination questions in iris-access form now use this) and `storageTarget` dropdown in Sanity Studio.
- Shares allowlist validation between Studio (`form-field-storage-validation`) and runtime mapper (`iris-access-allowlists-manifest`).
- Site form renderer, Zod schema, and GROQ updated to expand checkbox groups into flat boolean answer keys.

**Signup & onboarding**
- Approval emails use `/en/signup?application_id=&token=` only - no name/email/phone in query params.
- Password step prefill comes from server-loaded application answers (`informalName` -> farm name); server action falls back to DB if hidden fields are empty.
- Removed legacy `/incoming` route, redirect, and URL-param signup fallbacks.
- Consumed signup links redirect to `/apply` instead of showing “invalid or expired”.
- `/apply` slimmed (General Business / Farm tabs removed); authenticated routes gated until apply onboarding completes.

**Internal dashboard**
- Application detail: Standard + Technical tabs (raw JSON + DB hydration preview via `preview-form-submission-hydration`).
- Farm advisor profile notes editor on application detail.

## Checklist


vincenttodd:
Looks really good. 2 questions:

- is the other incoming (I think that the route) and micro service removed for inviting clients to create password/ onboard on the old signup flow.
- if /signup route is post-request form (beginning of the onboarding flow), it doesn't need internationalized. Otherwise its fine.

yousefalwahami:
> Looks really good. 2 questions:
> 
> * is the other incoming (I think that the route) and micro service removed for inviting clients to create password/ onboard on the old signup flow.
> * if /signup route is post-request form (beginning of the onboarding flow), it doesn't need internationalized. Otherwise its fine.

Yes, the old flow is removed. Before: users got a link like `/incoming` with their name and email in the URL,
now they get `/en/signup?application_id=123&token=abc` with no personal info in the URL.
We removed the `/incoming` page, also removed reading name/email from the URL on signup. `/auth/accept-invite` still exists for co-workers and I also kept the `go.toddagriscience.com/invite` for creators/partners. There's no separate microservice for this in the repo, it's all in the main site app.

Good catch, right now its `/en/signup` since approved email link there and the routing setup lets logged in and logged out users open the link without it bouncing out (redirects for [locale] and platform), it works fine as-is and was expecting that issue had something to do with the locale, **but** I went ahead and removed the international since its just unrelated.

longlong9889 (changes requested):
good PR! The core goal (PII out of signup URLs) is clean: the link carries only application_id + token, prefill is resolved server-side, and validation uses the stored
applicant email (not the URL/form). The form→farm hydration uses per-table column allowlists, so unknown answer keys can't inject columns. Token checks (approved + single-use via signedUpAt + expiry + soft-delete) and the rename migration both look right, and the schema table name matches the migration.

Two robustness issues in `completeFormSubmissionSignup`, both verified with a small mock-based test.

issue 1 — form-submission-signup.ts, on the await hydrateFarmFromFormSubmission(...) call

 Hydration runs after the account + farm exist and the row is already marked `signedUpAt`, with no try/catch. Answer values are key-allowlisted but never type-checked, and `total_acreage` is an `integer` column / `total_gross_income` is `numeric` — so a free-text answer (e.g. "around 120") makes Postgres throw, hydration throws, and `signUp` throws *after* the account was created and the token consumed. I think you can wrap hydration in try/catch + logger.error so a mapping error degrades to "not hydrated" instead of breaking signup.

 Verified with a mock of the db chains + hydrate helper: `completeFormSubmissionSignup` rejects when hydration throws, and `updateCalls === 1` before the throw — i.e. the row is already consumed, so the user ends up with an account + a consumed link *and* an error.

issue 2 — form-submission-signup.ts, on the db.update(...).set({ signedUpAt... }) block

This selects then updates with the same `isNull(signedUpAt)` guard, but hydration below isn't gated on the UPDATE actually flipping the row. Two concurrent calls both pass the initial SELECT and both hydrate (and hydrate's existence-check-then-insert is itself non-atomic, so farm_location/farm_certificate can double-insert). Maybe adding `.returning()` here and only hydrate when a row comes back.

Verified: two concurrent `completeFormSubmissionSignup` calls that both pass the initial SELECT both invoke `hydrateFarmFromFormSubmission` (asserted 2×), since the UPDATE rowcount is never checked.

longlong9889 (approved):
both issues fixed!

PR #925, opened 2026-06-11
feat(site): add sidebar UI to authenticated layout

C0derTang:
## What

Adds sidebar navigation to all authenticated routes in the `site` app and moves it into the authenticated layout.

## How

**Layout** (`apps/site/src/app/(authenticated)/layout.tsx`)
- Wraps authenticated content in `SidebarCollapseProvider` + `SidebarClient`, with the server-rendered `<Sidebar />` passed as children (keeps `Sidebar` a server component).
- Desktop: a width-animated panel that is sticky/pinned and collapses to zero width; a small expand button appears once collapsed. Mobile: a hamburger-triggered `Sheet` overlay.

**Sidebar components** (`apps/site/src/app/(authenticated)/components/sidebar/`)
- `sidebar.tsx` — server component: collapse control, primary nav links, a **read-only** management-zones list, and the account footer.
- `sidebar-client.tsx` — responsive wrapper (sticky desktop panel vs mobile sheet).
- `sidebar-collapse-context.tsx` / `sidebar-collapse-toggle.tsx` — persisted collapse state via `useSyncExternalStore` (localStorage) with a `getServerSnapshot`.
- `sidebar-nav-item.tsx` — nav link with `usePathname` active highlighting + `prefixMatch`.
- `sidebar-search-button.tsx`, `sidebar-user-footer.tsx`, `sidebar-section-label.tsx`, `sidebar-sheet.tsx`.
- `zone-item.tsx` — **read-only** zone row (links to the zone). No create/delete controls — adding/removing zones is a client-advisor action and is intentionally not exposed here.

**Icons**
- Rendered from the `react-icons/lu` library (`LuBell`, `LuShoppingCart`, `LuSearch`, `LuSettings`, `LuMenu`). No icon image files are committed — there is no `<Image>` icon wrapper and no `public/icons/*.svg`.

**Tests**
- `sidebar.test.tsx`, `sidebar-search-button.test.tsx`, `zone-item.test.tsx`, plus `sidebar-nav-item`, `sidebar-user-footer`, and `sidebar-collapse-toggle` tests (6 files / 16 tests).

## Notes
- The top-right nav (`AuthenticatedHeader`) being replaced by `AccountHeader` on account pages is a **pre-existing** inconsistency, out of scope here — tracked separately.

Closes the sidebar UI task.


vincenttodd:
top level, we need to add tests, and these are the tweaks that need to be done on the menu:

Screenshot 2026-06-11 at 11 26 07 AM

vincenttodd:
last couple things I just noticed: remove the option to add/delete zones, that can only be performed by a client advisor. Also import the icons from the react library. We don't use/keep icons files themselves. Otherwise visually LGTM

vincenttodd:
Oh and make sure theres testing where its needed

longlong9889 (changes requested):
I think its a nice refactor overall, moving the sidebar into the authenticated layout looks good, the server/client split is clean (async Sidebar passed as children intSidebarClient), and useSyncExternalStore with a getServerSnapshot is the correct way to handle the persisted collapse state. A few things I'd like to see resolved before approving.

issue #1 — attach to sidebar-client.tsx line 32

  **Sidebar `position scrolls away with the page.**

Confirmed locally: on an authenticated page tall enough to scroll the window, the sidebar slides up and out of view instead of staying pinned.

  Cause: this width-animation wrapper has `overflow-hidden`, and it wraps the sticky `<aside>` (`sticky top-0 h-screen` in `sidebar.tsx`). An ancestor with `overflow: hidden` becomes the sticky scroll-container, and since the wrapper is exactly as tall as the aside, there's no room to stick.

  Fix: make this wrapper itself the sticky element so stickiness resolves against the window:

  ```
        className={`sticky top-0 h-screen overflow-hidden transition-[width] duration-300 ease-in-out ${
  ```

`overflow-hidden` on the sticky element itself is fine (the rule only bites on *ancestors*), and the wrapper's parents set no overflow. The `sticky top-0 h-screen` on the inner `<aside>` then becomes redundant and can be dropped.

issue #2 — post on the PR description / top-level

The PR description is out of sync with the diff. It references an icon `<Image>` wrapper, `public/icons/*.svg` (9 files), `new-zone-button.tsx`, a `zone-item` with inline delete, and `createManagementZone` / `deleteManagementZone` actions which doesnt seem to be present in this PR. The actual implementation uses `react-icons` and a read-only zone list. Would you mind updating the description so the next reviewer isn't reviewing against the wrong spec please?

rcjasub (commented):
I think it looks great, it aligns with what vicent suggested. 
Just my two cents:

Top-right nav disappears on account pages
The AuthenticatedHeader (Contact, Account links) is replaced by AccountHeader on account pages, so the top-right nav vanishes entirely. Pre-existing but noticeable. worth making the nav consistent across all authenticated pages?

Dead code — sidebar-actions.ts
sidebarLogout was removed from the sidebar UI but the file wasn't cleaned up. Nothing imports or calls. 

Image
Image

longlong9889 (approved):
Looks good to me!

PR #927, opened 2026-06-11
feat(site): search modal (ported from local impl)

C0derTang:
## What

Creates the **search modal** fand makes search modal-only. The dedicated `/search` results page is removed; search now lives in the sidebar.

## Changes

**New**
- `apps/site/src/components/common/search-modal/search-modal.tsx` — client modal: Browse IMPs / Browse Seeds tabs, client-side `.filter()`, add-to-order toggle, Escape-to-close, backdrop dismiss.
- `apps/site/src/app/(authenticated)/actions/search-modal.ts` — `getSearchModalData()` server action (`Promise.all` of IMPs + seeds).
- `apps/site/src/lib/db/imps.ts` — `getAllImps()`.
- `apps/site/src/lib/db/seeds.ts` — `getAllSeedProducts()` (selected columns verified against the target `integratedManagementPlan` / `seedProduct` schema — no column drift).

**Modified**
- `sidebar-search-button.tsx` — now a `'use client'` component using `useState` + `<Portal>` + `<SearchModal>`, opened via `SidebarNavItem onClick` (no longer a `/search` link).
- `(widgets)/page.tsx` — removed `<SearchNavForm />` usage + import, and the `/search` empty-state nav item.
- `tabs.tsx` / `tabs.test.tsx` — removed dangling `SearchNavForm` render + mock from the (dead) tab view.
- `search-nav-form.tsx` — submit now returns to `/` (its `/search` results page is gone); live dropdown unchanged.
- `imp/[slug]/page.tsx`, `product/[slug]/page.tsx`, `order-client.tsx` — repointed `/search` links to `/`.

**Removed**
- The entire `apps/site/src/app/(authenticated)/(misc)/search/` route directory.

## Notes
- Grep for `/search` route references is clean repo-wide.
- `bun type-check` and `bun lint` (touched paths) pass; test suite green (the 2 flaky timeout failures in the full run are pre-existing, unrelated files, and pass in isolation).
- **This branch stacks on #925** (`feature/sidebar-ui`) and targets it as the base. **Retarget to `main` once #925 merges.**


rcjasub (commented):
Add to order" is UI-only right now, selections live in local component state and aren't wired to the order system, so navigating to /order shows an empty cart. Is this intentional for now?

C0derTang:
Good catch — it was intentional as a placeholder, but I've now wired it up properly in 80bf6db. The modal's "Add to order" uses the real `useOrder()` localStorage cart (same hook the product page uses) instead of throwaway local state, so seeds added from the modal now persist and show on `/order`. Added/footer state derives from the cart. Verified end-to-end: adding a seed in the modal shows it on `/order` with the correct qty/price. Added a search-modal add-to-order test.

rcjasub (commented):
LGTM on the core logic — useOrder() wired correctly, cart persists, focus management is solid.

 A few non-blocking notes (first two are pre-existing, not introduced here):
 - imps/seeds state is typed as any[] — worth adding proper types given the strict TS config.
 - No .catch() on getSearchModalData() — a fetch failure silently shows empty lists.
 - The "Added" button is still clickable and increments qty rather than removing — a toggle (Add ↔ Remove) would be better UX so users aren't stuck if they add something by mistake.

vincenttodd:
Couple things:

- reduce background shade, or just make the search box have a drop shadow. 

Screenshot 2026-06-18 at 3 28 51 PM

- I would allow users to still scroll behind the search module. 
- is it possible (or does it work like this) to allow user to search for both or just IMPs/Seeds? If the client is able to search both off the bat, it's confusing ngl.
- also probably skip the search button in the bottom right corner. Since there will be a window/module on the right side as well, it won't serve much purpose when the window is open.

C0derTang:
## Merge conflicts resolved + review items verified

### Merge `main` (`658ef02`)
Resolved three sets of conflicts:
- **`package.json`** (overrides) — took main's `undici@^7.28.0` (security patch) and kept main's added `nodemailer@^9.0.1`; `ws@^8.21.0` identical both sides.
- **`bun.lock`** — reset to main's lockfile, then `bun install` (reported no changes). This deliberately avoids the version drift that would otherwise bump `zod`/`@hookform/resolvers` and break type-check.
- **7 rename/delete conflicts** under `(authenticated)/.../search/` — this branch removed the standalone `/search` page (commit `4280dd8`, "add search modal and remove /search page"), while `main` only relocated those files into the `(shell)` group. Accepted the branch's deletion (`git rm`). Grepped for dangling `/search` links → none; `tsc --noEmit` clean.

### Review items — all addressed ✅
Verified each suggestion against the merged code:

| Reviewer | Suggestion | Status |
|---|---|---|
| @rcjasub | Wire "Add to order" to the real order system | ✅ `useOrder()` cart, `addItem`/`removeItem`, reads `order.items` |
| @rcjasub | `imps`/`seeds` typed (not `any[]`) | ✅ `Imp[]`/`Seed[]` from `Awaited<ReturnType<typeof getSearchModalData>>` |
| @rcjasub | `.catch()` on `getSearchModalData()` | ✅ `.catch()` → `loadError` + `role="alert"` message |
| @rcjasub | Add ↔ Remove toggle (not stuck after mis-add) | ✅ `aria-pressed`, click toggles remove, hover shows "Remove" |
| @vincenttodd | Lighter backdrop / drop shadow | ✅ `bg-black/20` + `shadow-2xl` |
| @vincenttodd | Allow scrolling behind the modal | ✅ no body scroll-lock (test asserts `body.style.overflow === ''`) |
| @vincenttodd | Search one type at a time (not both) | ✅ `activeTab: 'imps' \| 'seeds'` tabs |
| @vincenttodd | Drop the floating bottom-right search button | ✅ removed; trigger moved to `sidebar-search-button.tsx` (Portal-mounted) |

**Verification:** `tsc --noEmit` clean; search-modal + trigger tests pass (16 tests).

rcjasub (commented):
LGTM. Nice additions with the focus trap and race condition guard. (from the suggestion I made)

Two non-blocking items worth a follow-up:

- IMP expand row (~line 229) is a `div` — should be a `<button>` for keyboard access.
- No loading state while `getSearchModalData()` is in-flight — the list briefly shows "No IMPs found" before data arrives. A small spinner or skeleton would clean that up.

vincenttodd (approved):

PR #928, opened 2026-06-11
feat(site): reminders (ported from local impl)

C0derTang:
﻿> # WARNING — READ BEFORE MERGING: THIS PR MODIFIES THE PRODUCTION DATABASE
>
> ## DO NOT DEPLOY WITHOUT PROVISIONING THE `reminder` TABLE IN PROD
>
> **This feature adds a brand-new `reminder` table and a `reminder_type` enum. The schema file is the source of truth, but the live PRODUCTION database will NOT have this table until someone runs the migration against it.**
>
> **At deploy you MUST apply `packages/db/drizzle/0013_add-reminder.sql` (or the equivalent migration) to the production database. Until then, visiting the reminders route throws at runtime (relation "reminder" does not exist).**
>
> **Modifying the production database is irreversible and high-risk — coordinate with whoever owns prod DB access, take a backup first, and run it in a known-safe window. CI does NOT touch a live DB, so green checks here do NOT mean prod is ready.**

---

## What

Ports the **reminders** feature into `nightcrawler`. The sidebar already has a Reminders nav item (from #925); this adds the page and data layer behind it.

## Changes

- **Schema:** `packages/db/src/schema/reminder.ts` (+ barrel export); hand-written migration `packages/db/drizzle/0013_add-reminder.sql`.
- **Page:** `apps/site/src/app/(authenticated)/reminders/` — server `page.tsx`, `reminders-list.tsx`, `reminder-card.tsx`, `create-reminder-dialog.tsx`, `reminder-edit-sheet.tsx`, `reminder-form.tsx`, `actions.ts`, `db.ts` (user-scoped via `getAuthenticatedInfo`), `types.ts`, `parse-seasonal.ts`.
- **Tests:** `parse-seasonal.test.ts` + `reminders-list.test.tsx` (16 tests).

## Notes

- **Enum expanded to 7 types** so form options match the DB enum (the one intentional divergence from the legacy schema).
- **Stacks on #925** (base `feature/sidebar-ui`); retarget to `main` once #925 merges.

## Verify

`bun type-check` clean - `bun lint` clean on reminders - `bun run test` green (16 reminders tests). CI does not hit a live DB.

Generated with Claude Code


rcjasub (commented):
1. winter: 12 causes wrong year (parse-seasonal.ts:63)
new Date(year, 12, 1) overflows to January of the next year — "early winter" and "late winter" both land in 2027 instead of 2026. Fix is 11 for December or 0 for January depending on intent, but 12 is always wrong.

2. Exact date off-by-one in US timezones (reminder-form.tsx:85)
new Date(dueDateStr) where dueDateStr is YYYY-MM-DD parses as UTC midnight, so in any US timezone it shows as the day before. Fix: new Date(dueDateStr + 'T00:00:00').

3. Both "Seasonal Label" and "Exact Date" can be submitted together
No validation prevents filling in both fields. Both get saved to the DB, and on the card the seasonal label text renders independently from the effective date — so a user sees "mid March" as the label but a completely different exact date underneath. Either clear one field when the other is filled, or block submission if both are set.

C0derTang:
Thanks for the review — all three fixed in b8fa71f:

1. **winter:12 overflow** — `seasonMap` corrected to real season-start months (spring 2 / summer 5 / fall 8 / winter 11). `new Date(year, 12, 1)` no longer rolls into next January; the sibling seasons were each one month high too, so those are fixed as well.
2. **Exact-date off-by-one** — exact dates now parse as local midnight via `new Date(dueDateStr + 'T00:00:00')` (both create and edit paths), so they no longer shift to the previous day in US timezones.
3. **Both fields settable** — Seasonal Label and Exact Date are now mutually exclusive: filling one disables the other, plus a submit-time guard drops the seasonal label if both are somehow set, so the card never shows a label/date mismatch.

Added a season-month overflow guard to `parse-seasonal.test.ts` and a mutual-exclusion test in `reminder-form.test.tsx`. Verified end-to-end in the running app.

rcjasub (commented):
LGTM. Might be worth a UI/UX pass from @vincenttodd   before merging?

C0derTang:
## Merge conflicts resolved + reminders sidebar fix

Brought `main` into the branch and fixed the fallout. Three commits:

### 1. Merge `main` (`41cbb5f`)
Resolved two conflicts:
- **`package.json`** (overrides) — took main's `undici@^7.28.0` (the high-severity security patch) over the branch's `^7.24.0`, and kept main's added `nodemailer@^9.0.1`. `ws@^8.21.0` was identical on both sides.
- **`bun.lock`** — regenerated.

### 2. Pin `zod`/`@hookform/resolvers` (`d61c640`)
The lockfile regen in step 1 silently bumped `zod` 4.3.6 → 4.4.3 and `@hookform/resolvers` 5.2.2 → 5.4.0 (both within range). That pair broke `zodResolver` type inference (`phone: unknown` vs the schema's `string`) in `accept-form.tsx:36`, failing **Code Quality (type-check)** and **Build**. Reset `bun.lock` to main's pinned versions (`bun install` then reported no changes — the branch needs no new deps). `package.json` unchanged. Local `tsc --noEmit` clean.

### 3. Restore reminders sidebar (`552dc51`)
The reminders route lived at `(authenticated)/reminders/`, **outside** main's new `(shell)` route group whose layout renders the platform sidebar — so the page showed no sidebar. Moved the whole dir into `(shell)/(misc)/reminders/` alongside `order`/`imp`/`product`/`v`. The URL stays `/reminders` (route-group parens are ignored in the path), the sidebar nav link is unchanged, and all imports are relative so nothing else needed touching.

**Verification:** `tsc --noEmit` clean; reminders test suites pass (34 tests). `mergeable: MERGEABLE`.

vincenttodd (approved):

C0derTang:
### Step by step to ship to prod
1. Confirm secrets exist on the repo: `PROD_DATABASE_HOST`, `PROD_DATABASE_PORT`, `PROD_DATABASE_USER`, `PROD_DATABASE_PASSWORD`, `PROD_DATABASE_DATABASE`, `DATABASE_PEM_CERT`. (Already used by prior releases.)
2. Take a fresh backup / snapshot of the prod database. This step is irreversible; do not skip.
3. Merge this PR into `main`. Note: merge alone does **not** touch prod.
4. Create a **GitHub Release** (or run the workflow via `workflow_dispatch`). This triggers `release-database-compatibility-check.yml`.
5. The workflow runs in order:
   - `pg_dump` prod, restore into a throwaway CI Postgres.
   - `drizzle-kit push --config=drizzle-local.config.ts` against the clone. This is the safety check: it proves the push applies cleanly against a real copy of prod before touching prod.
   - `drizzle-kit push --config=drizzle-prod.config.ts` against **real prod**. Drizzle diffs `src/schema/*` vs prod and applies the DDL, creating the `reminder` table and `reminder_type` enum.
6. Watch the workflow logs. If the validation push (against the clone) fails, prod is never touched. Fix and re-cut.
7. After it goes green, smoke-test the `/reminders` route in prod. Before the push runs, that route throws `relation "reminder" does not exist`.

### Watch out
- For this PR the change is purely additive (new table + new enum), so `push` is safe.
- `drizzle-kit push` can be destructive on **renames or drops** (it may drop+recreate and lose data). Not a concern here, but keep it in mind for future schema changes.

PR #929, opened 2026-06-11
feat: serve English at root path, drop /en/ prefix

### LINKED ISSUE #852: Default `EN-US` to non-locale `.com/` instead of .`com/en/`
### Problem Statement

We want to serve our site in the US (en) with a non locale domain (`.com/`), while keeping other languages with the standard `.com/locale/` format. 

### Context

_No response_

### Constraints

_No response_

### Technical Details

_No response_

### User Story

_No response_

### Use Case

_No response_

rcjasub:
Description

Fixes #852

## Serve English at `/` instead of `/en/`

Changes `localePrefix` from `'always'` to `'as-needed'` in the next-intl config so English is served at the root path while Spanish stays at `/es/`.

### What changed

- Added permanent redirects for all legacy `/en/*` URLs so existing links and bookmarks don't break
- Fixed a redirect loop in auth middleware where `/` was simultaneously a protected route and the new marketing homepage, unauthenticated requests to `/` now fall through to the intl middleware instead of redirecting
- Updated all internal hardcoded `/en/` links (privacy, terms, logout redirects, canonical URLs, etc.)
- Extended `specialRedirect` to handle both prefixed (`/en/careers/externship`) and unprefixed (`/careers/externship`) externship routes
- Updated proxy tests (`auth`, `i18n`, `all-users-page`, `special-redirect`) to cover the new routing behavior

Checklist

- [x] You've used conventional commits

### DIFF
--- a/apps/site/next.config.ts
+++ b/apps/site/next.config.ts
@@ -160,17 +160,6 @@ const nextConfig: NextConfig = {
 
   redirects() {
     return [
-      // Redirect legacy /en/* URLs — English is now served without a locale prefix
-      {
-        source: '/en',
-        destination: '/',
-        permanent: true,
-      },
-      {
-        source: '/en/:path*',
-        destination: '/:path*',
-        permanent: true,
-      },
       // Legacy slug redirects for English
       {
         source: '/who-we-are/:path*',
--- a/apps/site/next.config.ts
+++ b/apps/site/next.config.ts
@@ -160,24 +160,52 @@ const nextConfig: NextConfig = {
 
   redirects() {
     return [
+      // Redirect legacy /en/* URLs — English is now served without a locale prefix
       {
-        source: '/:locale(en|es)/who-we-are/:path*',
-        destination: '/:locale/about/:path*',
+        source: '/en',
+        destination: '/',
         permanent: true,
       },
       {
-        source: '/:locale(en|es)/who-we-are',
-        destination: '/:locale/about',
+        source: '/en/:path*',
+        destination: '/:path*',
+        permanent: true,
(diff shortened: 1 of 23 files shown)


rcjasub:
The window.location.assign() in footer.tsx (line 100) is pre-existing and not part of this change. Happy to address it in a separate PR if needed.

C0derTang:
**Wrong-locale regression on legacy /en/* links**

 Files: next.config.ts, proxy/i18n.ts:68-77

The redirect /en/:path* → /:path* strips the explicit English prefix. The resulting unprefixed path then passes through the locale-detection block in handleI18nMiddleware, which re-derives language from cookie/accept-language. A Spanish-preferring client is redirected onward to the Spanish copy, discarding the user's explicit English intent. Under the previous localePrefix: 'always' config the /en/ prefix was preserved end-to-end.

Reproduction (production build):
1. Ensure NODE_ENV=production so localeDetection is active.
2. Clear the NEXT_LOCALE cookie.
3. Set browser/request header Accept-Language: es-ES,es.
4. Request https://<host>/en/about.
5. Observe redirect chain ends at /es/about instead of an English page.

Expected: English content served (at /about). Actual: Spanish (/es/about).

C0derTang (approved):
Looks good to me

vincenttodd:
how is the /en route being managed on the sitemap?

rcjasub:
> how is the /en route being managed on the sitemap?

the sitemap wasn't updated to reflect the new URL structure. I'll fix it in this PR

vincenttodd (approved):

PR #943, opened 2026-06-15
Feat/variety inventory

### LINKED ISSUE #942: Mirror the Variety Inventory spreadsheet into the database (seed catalog foundation)
### Problem Statement

  We want clients to be able to browse, compare, and request seed varieties (with a
  desired amount) and not purchase them. To do that, variety data needs to live in the
  application database so it's searchable and linkable to management zones.

  The catch: the real variety list (1000+ rows) lives in a Google Sheet that the team
  actively maintains, and we don't want to migrate that master list into the database
  and get locked in. As Vincent put it: *"If it's going in a DB it's staying there."*
  We also don't want to build and pay for heavy custom infrastructure to manage a list
  we can manage almost for free elsewhere.

  So the problem is: how do we get this spreadsheet's data into the database for the
  app to use, without making the database the source of truth and without a risky
  one-way migration?

 
### Context

  - The master list is a Google Sheet ("Variety Inventory") the team edits directly.
    It is hierarchical: ~270 crop groups (e.g. BASIL, BEET), each with nested
    varieties/cultivars (~940 total).
  - Per variety we need: name, description, growing info (to give the zone inference
    search agronomic context), display-only pricing, and a simple stock status
    (in stock / back order). Some rows are historical references only ("Cultivar
    Descrpt.") and are not requestable.
  - There is **no Stripe / no purchasing** in scope — clients request varieties, they
    don't buy them. The existing `seed_product` commerce tables are a separate concern.
  - A `crop` table already exists for a different (zone-planting) feature, so the
    catalog tables must use distinct names.

_No response_

### Constraints

_No response_

### Technical Details

_No response_

### User Story

_No response_

### Use Case

_No response_

longlong9889:
Fixes #942
## Description
Adds a **regenerable database mirror** of the "Variety Inventory" Google Sheet so the app can browse/compare/request seed varieties without making the database the source of truth. The Sheet stays the master; this just rebuilds the DB copy on demand.

## What's included

  - `seed_crop` + `seed_variety` schema and migration `0012`
  - Re-runnable, idempotent CSV importer (`packages/db/scripts/import-seed-varieties.ts`,
    script `db:import-varieties`): real CSV parsing, crop/variety classification,
    auto-derived request status, price→cents parsing, slug generation, note-skipping,
    and content-hash–cached embeddings
  - Localhost safety guard (refuses to run against anything but the local DB)

## How to test

  ```bash
  # export the Sheet to packages/db/data/variety-inventory.csv, then:
  cd packages/db
  bunx tsx scripts/import-seed-varieties.ts            # dry run
  bunx tsx scripts/import-seed-varieties.ts --commit   # writes to local DB
  Expected: ~270 crops, ~943 varieties, status reference=28, orphans 0.

  Notes for reviewers

  - Embeddings are deterministic placeholders locally (no OPENAI_EMBEDDINGS_KEY). Real vectors are a follow-up once the key is available.


yousefalwahami:
Good PR, schema naming, status model, and localhost guard all look right, and the importer matches our existing `knowledge_article` pattern.

One question before merging: the PR describes a “rebuild,” but the script is upsert-only. Meaning It never removes rows that were deleted from the sheet, if someone removes a variety or renames a crop, the old DB row will stick around. Is this okay or do we want a cleanup in this PR/follow-up?

Nits:
- Small typo in the PR it says migration `0010` but the branch adds `0012`. 
- Consider a one-liner in `packages/db/README.md` for the CSV export + `db:import-varieties` workflow.

longlong9889:
I think the bun audit failure is unrelated to this PR. The branch changes zero dependencies, the only package.json edit is adding the db:import-varieties script, and git diff main...HEAD -- bun.lock is empty, so the dependency tree is byte-for-byte identical to main.

Running bun audit --audit-level=high against that same lockfile reproduces the exact same 4 highs:

  - form-data (via sanity)
  - ws (via openai / storybook / supabase / sanity / newrelic)
  - protobufjs (via newrelic / posthog-js)
  - vite (via storybook / vitest / sanity)

All are pre-existing transitive deps. They surface now maybe because bun audit checks the live advisory DB and these CVEs were published after the lockfile was last pinned, so main and every open PR fail the same gate today, regardless of their diff.

This belongs in a separate dependency-bump chore rather than a schema PR. Is this check blocking for merge, or can it be handled separately?

yousefalwahami (approved):

PR #886, opened 2026-05-14
Refactor/careers page

yousefalwahami:
## Description

Refactor the careers marketing hub at `/careers`, including new job listings being displayed at `/careers/index` with filters, reusable marketing blocks and assets, also including `careers/[job-slug]` for jobs, no longer using `index/[job-slug]`, Sanity `career/schema` and article handling updates, and marketing route renames (about, research, legal) with nav, footer, and sitemap updates.

## Checklist

- [x] You've included unit or integration tests for your change, where applicable.
- [x] You've included inline docs for your change, where applicable.
- [x] Any components that you've modified are accessible.
- [x] You've used [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/) where appropriate

PR #944, opened 2026-06-15
feat(careers): replace exposed accommodation email with inquiry link

### LINKED ISSUE #590: Add an inquiry link to /career disclosure footer for disability accommodation
### Problem Statement

Current there is a live email address publicly available in the career page when someone needs to request a disability accommodation.

### Context

Compliance

### Constraints

_No response_

### Technical Details

_No response_

### User Story

A applicant should be able to open a inquiry ticket to request a disability accommodation

### Use Case

_No response_

BaoT1301:
## Description

Fixes #590

### What
The careers disclosure footer (disclaimer item #2) listed `support@toddagriscience.com` in plain text as the channel for disability accommodation requests. This replaces it with a locale-aware link to the existing `/contact` public inquiry form.

### Why
- A live email address in page copy is exposed to scrapers and offers applicants no structured way to open a request.
- The issue asks for "an inquiry link ... so an applicant can open an inquiry ticket." `/contact` already creates a public inquiry ticket, so we route there instead of exposing the mailbox.

### How
- Added an optional `linkHref` prop to the shared `Disclaimer` component. When provided, disclaimer strings may use a `<link>...</link>` tag rendered via next-intl's `t.rich`; the careers page passes `linkHref="/contact"`.
- Callers that omit `linkHref` (terms, research) keep the original plain-text rendering, so the change is fully backwards compatible.
- Updated both `en` and `es` accommodation strings to use the `<link>` tag.

### Testing
- `careers` + `disclaimer` suites pass (11 tests); type-check, lint, and build all green.

## Checklist

- [x] You've included unit or integration tests for your change, where applicable. _(covered by existing careers/disclaimer suites; behavior is backwards compatible)_
- [x] You've included inline docs for your change, where applicable. _(JSDoc on the new `linkHref` prop)_
- [x] Any components that you've modified are accessible. _(link has descriptive visible text + underline; no raw email)_

### DIFF
--- a/apps/site/src/app/(unauthenticated)/[locale]/(marketing)/careers/components/careers-externship.tsx
+++ b/apps/site/src/app/(unauthenticated)/[locale]/(marketing)/careers/components/careers-externship.tsx
@@ -115,6 +115,7 @@ export function CareersExternship() {
             <Disclaimer
               translationLoc="careers.disclaimers"
               disclaimerCount={5}
+              linkHref="/contact"
             />
           </div>
         </main>
--- a/apps/site/src/components/common/disclaimer/disclaimer.tsx
+++ b/apps/site/src/components/common/disclaimer/disclaimer.tsx
@@ -1,21 +1,31 @@
 // Copyright © Todd Agriscience, Inc. All rights reserved.
 
+import { Link } from '@/i18n/config';
 import { useTranslations } from 'next-intl';
 
 /**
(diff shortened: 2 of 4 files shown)


BaoT1301:
Closing in favor of #946. Investigation showed the careers refactor (#886) already removed the disclaimer (and the exposed email) from the live site — this branch edited the now-orphaned `careers-externship.tsx`, which never renders. #946 takes the correct approach: deleting the dead component + unused strings and closing #590 as already-resolved.

PR #945, opened 2026-06-15
fix(wordmark): preserve locale on home navigation

### LINKED ISSUE #940: [Bug]:  Locale not preserved on wordmark and login navigation
### Bug Description

Several internal links hardcode / or /en instead of preserving the user's current locale. When browsing in Spanish (/es) and clicking the Todd logo or the login link, the user is redirected to the English version of the page.

### Steps to Reproduce

1. Navigate to /es
2. Click the Todd wordmark — lands on / (redirects to /en)
3. Click Login in the header — lands on /en/login instead of /es/login

### Expected Behavior

All navigation should preserve the user's current locale.

### Screenshots/Videos

_No response_

### Browser

Chrome

### Device Type

Desktop

### Language/Locale

Spanish (es)

### Additional Context

_No response_

BaoT1301:
## Description

Fixes #940

### What
Clicking the Todd wordmark while browsing in a non-default locale (e.g. `/es`) sent the user back to the default-locale (English) home. The wordmark now preserves the current locale.

### Why
- The shared `ToddHeader` used `next/link` with `href="/"`. Because English is served at the root (`localePrefix: 'as-needed'`), `/` resolves to English, so the locale was lost on every logo click in Spanish.

### How
- Added an opt-in `localeAware` prop to `ToddHeader`:
  - **`localeAware`** → renders the locale-aware `Link` from `@/i18n/config`, which prefixes non-default locales (`/es/`). Applied to the localized headers/footer: new-header desktop + mobile nav, marketing footer, and the unauthenticated/contact header.
  - **default (omitted)** → renders a plain `next/link` (unchanged behavior).
- The default stays a plain link **on purpose**: the authenticated platform and `(go)` route groups have no `NextIntlClientProvider`, so rendering the locale-aware `Link` there (e.g. the client-rendered account header) would throw a missing-context error. Those headers are unlocalized (home is always `/`) and keep the plain link.

### Testing
- header / tabs / footer / nav / wordmark suites pass (45 tests across 15 files); type-check, lint, and build all green.
- Manually verified on `/es`: wordmark (desktop, mobile, footer) keeps the user on `/es`.

## Checklist

- [x] You've included unit or integration tests for your change, where applicable. _(covered by existing header/tabs/footer/nav/wordmark suites — 45 passing)_
- [x] You've included inline docs for your change, where applicable. _(JSDoc on the new `localeAware` prop explaining the provider constraint)_
- [x] Any components that you've modified are accessible. _(`aria-label` and `data-testid` preserved)_


rcjasub (approved):
Wordmark fix completes the issue. LGTM.
Approved the code changes but holding off on final merge — Security Audit check is failing. Can you look into what it's flagging?

PR #948, opened 2026-06-18
fix(internal): authorize every dashboard server action

### LINKED ISSUE #931: No auth checks on internal dashboard server actions
## Issue

`getUsers`, `createUser`, and other internal dashboard server actions rely entirely on middleware with no checks inside the actions themselves.

## Fix

Add a `requireInternalAccount()` guard at the top of every action and throw if it returns null. Middleware alone is insufficient — server actions can be called directly.

BaoT1301:
## Description

Closes #931

The internal dashboard only enforced auth in middleware. Problem is that `'use server'` actions are real POST endpoints you can hit directly, so middleware isn't a real boundary on its own — someone could call `getUsers`, `deleteFarm`, `approvePlatformAccessApplication`, etc. without being an internal member.

This adds a small `requireInternalAccount()` helper (`apps/internal/src/lib/require-internal-account.ts`) that checks the Supabase session against the `internal_account` table and throws if the caller isn't an active member, then calls it as the first line of all 47 actions across the 8 dashboard action files. I put it before each `try/catch` on purpose, so an auth failure throws clearly instead of getting swallowed and logged as a normal "operation failed".

It uses the same lookup that's already in `proxy/auth.ts` so middleware and the actions agree on who counts as a member. While I was in there I also refactored the existing `getReviewerAccountId()` in `applications/actions.ts` to reuse the helper instead of duplicating that query.

Things I left alone on purpose: the middleware (it works, and it has that sensitive "don't add logic between createServerClient and getClaims" bit), and I didn't wrap everything in some auth HOC/RBAC layer — explicit calls are easier to read and review. Kept `getClaims()` to match the existing code rather than swapping to `getUser()`.

On testing: `apps/internal` doesn't have a test harness yet and CI only runs the site tests, so I didn't add a unit test here. I did check it live on the local dashboard though — logged in as an active member and everything loads normally, then flipped the account to inactive and it bounced me to login. type-check, build, and lint all pass.

## Checklist

- [x] You've included unit or integration tests for your change, where applicable. (no harness in apps/internal yet — verified live instead)

### DIFF
--- a/apps/internal/src/app/(dashboard)/actions.ts
+++ b/apps/internal/src/app/(dashboard)/actions.ts
@@ -6,6 +6,7 @@ import { db } from '@nightcrawler/db';
 import { internalAccount } from '@nightcrawler/db/schema';
 import { eq, ilike, or, sql } from 'drizzle-orm';
 import logger from '@/lib/logger';
+import { requireInternalAccount } from '@/lib/require-internal-account';
 
 /**
  * Fetches all internal accounts, optionally filtering by search query.
@@ -13,6 +14,7 @@ import logger from '@/lib/logger';
  * @returns Array of internal account records
  */
 export async function getInternalAccounts(search?: string) {
+  await requireInternalAccount();
   try {
     if (search) {
       return await db
@@ -46,6 +48,7 @@ export async function createInternalAccount(data: {
   title?: string;
   isActive?: boolean;
 }) {
+  await requireInternalAccount();
   try {
     const [account] = await db.insert(internalAccount).values(data).returning();
     return account;
@@ -71,6 +74,7 @@ export async function updateInternalAccount(
     isActive?: boolean;
   }
 ) {
+  await requireInternalAccount();
   try {
     const [account] = await db
       .update(internalAccount)
@@ -90,6 +94,7 @@ export async function updateInternalAccount(
  * @returns True if deleted successfully
  */
 export async function deleteInternalAccount(id: number) {
+  await requireInternalAccount();
   try {
     await db.delete(internalAccount).where(eq(internalAccount.id, id));
     return true;
--- a/apps/internal/src/app/(dashboard)/analyses/actions.ts
+++ b/apps/internal/src/app/(dashboard)/analyses/actions.ts
@@ -6,6 +6,7 @@ import { db } from '@nightcrawler/db';
 import { analysis, mineral } from '@nightcrawler/db/schema';
 import { eq, ilike, or } from 'drizzle-orm';
 import logger from '@/lib/logger';
+import { requireInternalAccount } from '@/lib/require-internal-account';
 
 /**
  * Fetches all analyses, optionally filtered by search query.
@@ -13,6 +14,7 @@ import logger from '@/lib/logger';
  * @returns Array of analysis records
  */
 export async function getAnalyses(search?: string) {
+  await requireInternalAccount();
   try {
     if (search) {
       return await db
@@ -39,6 +41,7 @@ export async function getAnalyses(search?: string) {
  * @returns Array of mineral records
  */
 export async function getMineralsForAnalysis(analysisId: string) {
+  await requireInternalAccount();
   try {
     return await db
       .select()
@@ -79,6 +82,7 @@ export async function createAnalysis(data: {
     actionableInfo?: string;
   }>;
 }) {
+  await requireInternalAccount();
   try {
     const [newAnalysis] = await db
       .insert(analysis)
@@ -124,6 +128,7 @@ export async function updateAnalysis(
     macroActionableInfo?: string;
   }
 ) {
+  await requireInternalAccount();
   try {
     const updateData: Record<string, unknown> = {};
     if (data.analysisDate)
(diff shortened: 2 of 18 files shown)


BaoT1301:
Gave this a live test on the local internal dashboard with a throwaway internal account (deleted it after).

Logged in as an active member and the dashboard loaded fine with all its data, so the guards don't get in the way of normal use. Then I set that account's `is_active` to false and reloaded — got bounced to the login page, so a non-member is blocked.

type-check, build, and lint are green too.

rcjasub (dismissed):
Overall: The core approach is correct — requireInternalAccount() is well-structured, the placement before each try/catch is intentional and sound, and centralizing the lookup is the right call. Two things need fixing before merge.

---
approvePlatformAccessApplication / rejectPlatformAccessApplication — silent null audit record
apps/internal/src/app/(dashboard)/applications/actions.ts · lines 98–100, 173–175

Both actions call requireInternalAccount() twice — once as the guard (result thrown away), then again inside getReviewerAccountId(). If that second call fails for any reason, getReviewerAccountId() returns null and neither action guards on that before writing to the DB, so reviewedByInternalAccountId: null gets saved silently.

Fix — just use the return value from the first call:
// before
await requireInternalAccount();
const reviewerId = await getReviewerAccountId();

// after
const { id: reviewerId } = await requireInternalAccount();
You can then delete getReviewerAccountId() entirely.

---
Client components missing catch — unhandled rejections on auth failure
application-detail-client.tsx · handleApprove / handleReject / handleResendInvite
farms-client.tsx · load() / onSubmit / handleDelete

Before this PR, action failures returned null and handlers dealt with that. Now requireInternalAccount() throws before the action's try/catch, so the client handlers need a catch — they currently only have finally or bare calls. Without it, an expired session mid-action causes an unhandled rejection with no toast and no user feedback.

---
Minor — reviewerId fetched in updateFarmAdvisorProfileNotes but never written to the DB
apps/internal/src/app/(dashboard)/applications/actions.ts · lines 240–252

db.update(farm).set({ advisorProfileNotes: notes }) doealAccountId. Either the column is missing from the UPDATE or the getReviewerAccountId() call can be removed.

BaoT1301:
Thanks @rcjasub — all three addressed in the latest commit:

1. **Silent null audit** — `approve`/`reject` now take the account from the top-level guard (`const { id: reviewerId } = await requireInternalAccount()`) and use that for `reviewedByInternalAccountId`, so it can't be written null. Deleted `getReviewerAccountId()` since it's no longer needed.

2. **Missing catch in client handlers** — added a small shared `notifyActionError()` helper and wrapped the action calls in a try/catch. I did the two you flagged (application-detail, farms) and also the rest that had the same bare pattern (analyses, users, imps, seed-products, internal-accounts, tabs-widgets), since they'd hit the same unhandled rejection on an expired session. An auth failure now shows a toast instead of failing silently.

3. **Unused reviewerId in updateFarmAdvisorProfileNotes** — removed the lookup. It was fetched but never written, and the guard already covers authorization. Didn't add an editor column to the UPDATE since that'd be a schema change; happy to file that separately if we want to track who edits notes.

One note: locally the internal `build` hits an unrelated `/_global-error` prerender error (`useContext` null) — it reproduces on the base commit without these changes too, so it's environmental on my end. type-check + lint pass, and CI builds it fine.

BaoT1301:
Hey @rcjasub, thanks again for the thorough review, those were good catches. I went ahead and fixed all three: approve/reject now use the account from the guard call so the audit field can't be null, I dropped the unused reviewerId in the notes action and I added a little shared notifyActionError() helper and wrapped the action calls with a catch across all the dashboard client components (not just the two flagged ones, since they all had the same gap).

One thing though, the Security Audit is failing now but it doesnt look related to my changes. It is flagging two newly disclosed high severity CVEs in nodemailer and undici, both of which are existing deps I didnt touch. Everything else (build, tests, code quality, Vercel) is green.

For this should i create a separate fix(deps) PR to bump those two and clear the audit for everyone, or is someone already on it / would you rather Dependabot handle it? Happy to take it either way! Thank you so much!

rcjasub (commented):
LGTM — all three issues are cleanly resolved. The guard placement, the audit field fix, and the notifyActionError() coverage across all client components are all correct.

One thing outside this PR's scope but worth a follow-up: deleteInternalAccount lets any active member delete any other member's account — there's no role check or self-deletion guard. The requireInternalAccount() return value is discarded so there's no way to compare the caller's id against the target. Not a blocker here since this PR's job was guarding the actions, not defining what authorized members can do to each other — but it should be a tracked issue. @PratGit1606

On the security audit failure — before filing a separate fix(deps) PR , rebase on main first. PR https://github.com/toddagriscience/nightcrawler/pull/947 already patched several high-severity advisories via overrides and may cover these two as well. If the audit still fails after rebasing, a separate fix(deps) PR for nodemailer and undici is maybe the right move @vincenttodd

BaoT1301:
Thanks @rcjasub! I tried the rebase but my branch already has the latest main, so nothing new to pull. I looked into #947  overrides too but they don't cover these two: nodemailer isn't in there, and the undici override is still vulnerable so the audit is still red.

I think a separate fix(deps) PR is the move, I will bump undici to ^7.28.0 and nodemailer above 9.0.0. nodemailer's a major bump so I will test the email flow first. @vincenttodd, i was wondering if thats okay? or lmk if someone else should take it, otherwise I will start it!

vincenttodd (approved):

PR #949, opened 2026-06-18
Title: fix(site): resolve chunk error and logout navigation

### LINKED ISSUE #848: [Bug]: Chunk error after logging out and/or navigating to `/news` page
### Bug Description

Logging out and/or navigating to the `/news` page results in a chuck error or [react error #418](https://react.dev/errors/418?args%5B%5D=text&args%5B%5D=).

### Steps to Reproduce

navigate to `/news` page

### Expected Behavior

Should not have a error

### Screenshots/Videos

_No response_

### Browser

Safari

### Device Type

Desktop

### Language/Locale

English (en)

### Additional Context

_No response_

rcjasub:
## Description

Fixes #848

Resolves two related issues reported on the `/news` page and after logout:

- **Logout error**: `auth-client.ts` was calling `redirect('/')` from `next/navigation`, which is server-only. Calling it from a client context throws an unhandled error that triggers the React error boundary. Replaced with `window.location.replace('/')` which works correctly in the browser and is more appropriate for logout since it clears all client state.
- **Incorrect page directive**: `news/page.tsx` had a `'use server'` directive which marks exports as Server Actions rather than a Server Component. Removed — pages are Server Components by Next.js convention.
- **ChunkLoadError**: After a new deployment, chunk filenames change. Clients with cached HTML fetch stale chunk URLs that 404, triggering React error #418. Added a ChunkLoadError check in `app/error.tsx` that reloads the page once when detected. A `sessionStorage` flag prevents an infinite reload loop if the error persists.

## Checklist

### DIFF
--- a/apps/site/src/app/(unauthenticated)/[locale]/(marketing)/news/page.tsx
+++ b/apps/site/src/app/(unauthenticated)/[locale]/(marketing)/news/page.tsx
@@ -1,7 +1,5 @@
 // Copyright © Todd Agriscience, Inc. All rights reserved.
 
-'use server';
-
 import { FeaturedNewsCarousel } from '@/components/common/news/featured-news-carousel';
 import { LatestNewsTable } from '@/components/common/news/latest-news-table';
 import {
--- a/apps/site/src/app/error.tsx
+++ b/apps/site/src/app/error.tsx
@@ -18,6 +18,25 @@ export default function Error({
   error: Error;
   reset: () => void;
 }) {
+  // ChunkLoadError fires when the browser fetches a JS chunk that no longer
+  // exists after a new deployment. Reload once to pick up fresh chunks.
+  // The sessionStorage flag prevents an infinite reload loop if the error persists.
+  if (
+    error.name === 'ChunkLoadError' ||
+    error.message?.includes('Loading chunk')
+  ) {
+    try {
+      if (!sessionStorage.getItem('chunk-reload-attempted')) {
+        sessionStorage.setItem('chunk-reload-attempted', '1');
+        window.location.reload();
+        return null;
+      }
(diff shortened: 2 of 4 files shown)


Jose-Gael-Cruz-Lopez:
Closing in favour of #1088, which rebases this onto `main` and completes the cutover.

Your diagnosis was exactly right and the bug is **still live on main**: `auth-client.ts` is a `'use client'` module that ends `logout()` with `redirect('/')` from `next/navigation`. `redirect()` signals by throwing, so from a click handler `logout()` rejects instead of returning and the user is signed out but stranded. `window.location.replace('/')` is the right fix and #1088 uses it.

Two reasons to land #1088 instead:

**1. The callers.** Making `logout()` hard-navigate is only safe if every caller stops routing too. There are exactly 4 on main — `account-side-menu`, `auth-error-fallback`, `desktop-gate`, `logout-link` — and all 4 currently do `router.push('/')` (`logout-link` even carries the comment *"If logout doesn't redirect automatically, manually redirect"*). This PR changes `logout()` and no callers, which would ship 4 live races between a client push and a document navigation. #1088 removes the push from 4/4.

**2. The `'use server'` hunk is already done.** `news/page.tsx` was rewritten on main to render `ArticleIndex` and the directive is gone — and that obsolete hunk is the sole cause of this PR's merge conflict.

On chunk recovery: #1088 also has it, using a 60s time-windowed guard in an effect with a `role="status"` affordance and 155 lines of tests. This branch does `sessionStorage.setItem` + `window.location.reload()` **during render** and returns `null`, which isn't safe under StrictMode/concurrent rendering, and its guard is sticky — once set it persists for the session, so the next deployment's chunk error shows the error UI instead of recovering.

PR #954, opened 2026-06-19
feat: variety discovery page (/v) browse + detail

### LINKED ISSUE #951: Build client-facing variety discovery page (/v)
### Problem Statement

Clients have no way to browse Todd's seed inventory in the platform. The variety data now exists in the database (the seed_crop / seed_variety mirror of Vincent's "Variety Inventory" sheet — 270 crops / 943 varieties), but there is no UI on top of it. We need a client-facing page where authenticated users can browse, filter, and compare varieties,
view a variety's details and growing info, and request a variety + amount. There is no purchasing/checkout, clients browse and request only; the sheet remains the source of truth and the DB is a regenerable mirror.

### Context

- Data layer is done. seed_crop and seed_variety are in the schema, with a re-runnable importer (packages/db/scripts/import-seed-varieties.ts) that rebuilds the mirror from a CSV export of the sheet. Each variety carries: name, description, status (available / back_order / reference), display-only pricing (pricePerOzCents / pricePerLbCents / pricePerPlantCents), inventory note, last produced, location, and a FK to its parent crop (which holds the growing description). Embeddings for each crop/variety already live in the shared knowledge_article table.
- No images. The data is text-only by design (per Vincent, modeled on vercel.com/templates and vercel.com/blog/category/engineering).
- Design direction approved. A static preview of the browse layouts (list + grid) was reviewed and greenlit by Vincent. List groups by crop; grid is a card layout. Both use status badges (in stock / back order / reference) and a "price-from" summary.
- Route confirmed: /v for browse, /v/[slug] for detail (Vincent's call on the short prefix).

### Constraints

- Read-only over the mirror. The page reads seed_crop / seed_variety but must never write to or edit variety data — the Google Sheet is the source of truth and the DB is a regenerable mirror (the importer rebuilds it). No "edit variety" affordances in the UI.
- No purchasing / no Stripe. This is browse + request only. Pricing is display-only (stored as integer cents → format for display; never sum, never charge). Do not touch the existing seed_product / checkout flow.
- reference status is not orderable. Varieties with status reference are historical descriptions — hide pricing and the request action for them; show a "historical reference" treatment.
- Text-forward, no images. The schema has no image field by design. Layout must work entirely on typography/metadata (per the approved Vercel-style direction); don't introduce image placeholders.
- Inherit the shell — don't rebuild it. Render page content only. The authenticated sidebar + app-shell layout already comes from (authenticated)/layout.tsx; rendering your own header/sidebar will double up (this is exactly what broke PR #915). Gate the page with requirePlatformOnboardingComplete() like sibling routes.
- Follow the existing page pattern. Mirror product/[slug]/ (page.tsx + utils.ts + components/ + types.ts) for structure and DB-access conventions, so it's consistent with what's already there.
- Search stays light in v1. The knowledge_article embedding is vector(3072), which exceeds pgvector's index limits — there's no vector index, so semantic queries are sequential scans. v1 should use a simple text filter; defer semantic search rather than ship an unindexed vector query on the browse path.
- Slug is the routing key. /v/[slug] resolves by seed_variety.slug (unique, stable, generated by the importer). Unknown slug → notFound() (404), matching /product/[slug].
 - Volume. ~270 crops / ~943 varieties — the browse page should render that many rows comfortably (grouping/lazy-rendering as needed), not assume a tiny list

### Technical Details

_No response_

### User Story

_No response_

### Use Case

_No response_

longlong9889:
## Description

Fixes #951 

Variety discovery page (/v)

Adds the client-facing browse + detail UI on top of the seed_crop / seed_variety data layer from #943. Clients can now browse the seed inventory, filter and search it, and view a variety's details. Browse and request only — no purchasing/checkout.

What's included

  - /v — browse page: all varieties grouped by crop, with a search box, status filter (All / In stock / Back order / Reference), and a list ↔ grid layout toggle. Both layouts were reviewed and approved by Vincent.
  - /v/[slug] — detail page: variety description, parent crop's growing info, full price table (oz/lb/plant), inventory/last-produced/location, and a "Request this variety" CTA.
  - Status badges map to the seed_variety_status enum: available → In stock, back_order → Back order, reference → Reference.

Key decisions

  - List is the default layout, grid is a toggle (both ship). Closest match to the design references and handles the ~270-crop grouping cleanly.
  - reference varieties hide pricing and the request CTA — they're historical records, not requestable.
  - Pricing is display-only (cents → formatted); never summed or charged. The existing seed_product/checkout flow is untouched.
  - Request CTA links to /contact ("Request through an advisor") rather than a form — the request backend (where a request is stored) is an open question, deliberately deferred.
  - Search is a simple text filter (name / crop / description). Semantic search is deferred (the vector(3072) embedding has no pgvector index).
  - Split into utils.ts (server, owns the DB query) vs format.ts / types.ts (pure, client-safe) so the client component never pulls the DB connection into the browser bundle. Page is a Server Component; only the interactive browse component is 'use client'.

Not included (follow-ups)

  - No sidebar link yet — reviewers need to navigate to /v by URL for now. Nav entry to come separately.
  - Request-storage backend, semantic/zone-aware search, variety images.

Testing

  - Verified locally against the local Docker DB (270 crops / 943 varieties imported): browse renders grouped varieties, search/status filters and list/grid toggle work, detail pages resolve by slug, unknown slug → 404, reference varieties hide price + request.


yousefalwahami:
Nice PR - clean server/client split, good alignment with `/product` and `/imp` patterns, and the scope matches well

One thing to double-check: reference varieties hide pricing on `/v/[slug]` but still show `from $X / unit` on the browse list/grid. If reference rows shouldn't expose pricing anywhere, gate `priceFromLabel` when `status === 'reference'`.

Nits/follow-ups (non-blocking):
- Add `v/[slug]/layout.tsx` + `generateMetadata` for variety name in the tab title (like `/product/[slug]`)
- a11y: label the search input; `aria-pressed` on filter/view toggles

yousefalwahami (approved):

PR #955, opened 2026-06-20
fix(site): stop build-storybook EEXIST race (disable Vite publicDir copy)

C0derTang:
## Problem
`build-storybook` intermittently fails with:
```
EEXIST: file already exists, mkdir './storybook-static/fonts'
  at async mkDirAndCopy (node:internal/fs/cp/cp)
  at async copyDir
  at async Promise.all (index 1)
```

It's a **race condition**, not a deterministic failure:

- `staticDirs: ['../public']` makes Storybook copy the `public` dir into `storybook-static/`.
- The `@storybook/nextjs-vite` framework runs on Vite, and **Vite also copies its own `publicDir`** (defaults to `public`) into the build output.
- So the same directory is copied **twice, concurrently** (note `Promise.all (index 1)` in the trace). Both copies walk `public/`, and both call `mkdir storybook-static/fonts`.
- `node:fs/cp` does mkdir-then-copy. If copy A creates `fonts/` and copy B calls `mkdir fonts/` while it already exists → `EEXIST`. If the two happen to interleave the other way, it passes.

Because the outcome depends on scheduling/disk/file-count/machine load, it **fails some fraction of runs and passes others** — reliably enough to flake CI (shared, loaded runners), rarely on a fast idle laptop. An earlier change removed an overlapping `../public/fonts` `staticDirs` entry, but this Vite-vs-`staticDirs` double-copy of `../public` remained.

## Reproducing it
The flakiness makes a single run unreliable; run it in a loop (without the fix) and it fails a fraction of the time — more often under load / on slower disks / in CI:

```bash
cd apps/site
# without the fix on this PR:
for i in $(seq 1 20); do
  rm -rf storybook-static
  bun run build-storybook 2>&1 | grep -q EEXIST && echo "run $i: EEXIST" || echo "run $i: ok"
done
```

You'll see `EEXIST` on some fraction of iterations. It surfaces most reliably in CI (the original report came from a CI `build-storybook` job). With the fix applied, every iteration passes.

## Fix
Set Vite's `publicDir` to `false` in `viteFinal` so only Storybook's `staticDirs` copy runs — one writer, no race. This is Storybook's documented guidance for Vite-based frameworks.

```ts
viteFinal: async (config) => {
  config.publicDir = false;
  // ...aliases
  return config;
}
```

This removes the second copy entirely, so the collision is **impossible** (not just unlikely). One file changed: `apps/site/.storybook/main.ts`.

## Verification
Ran `bun run build-storybook` repeatedly (3× consecutive) — completes every time with no `EEXIST`. `public` assets (e.g. `/fonts`) are still served via `staticDirs`.

> Split out of #950 (sidebar work) so it can land independently — the race affects every branch, not just that PR.

### DIFF
--- a/apps/site/.storybook/main.ts
+++ b/apps/site/.storybook/main.ts
@@ -26,6 +26,12 @@ const config: StorybookConfig = {
   staticDirs: ['../public'],
   viteFinal: async (config) => {
     const rootDir = path.resolve(__dirname, '..');
+    // Storybook already copies `../public` via `staticDirs`. Vite ALSO copies
+    // its `publicDir` (defaults to `public`) into the build output, so the same
+    // directory is copied twice concurrently and node:fs/cp races on mkdir
+    // (intermittent `EEXIST` on storybook-static/fonts during build-storybook).
+    // Disable Vite's copy so only the staticDirs copy runs.
+    config.publicDir = false;
     // Add support for local fonts and path aliases matching tsconfig
     if (config.resolve) {
       config.resolve.alias = {


rcjasub (approved):
LGTM. Ran build-storybook 5 times locally — all clean, no EEXIST. Fix is correct: disabling Vite's publicDir in viteFinal removes the second copy entirely so the race can't happen. The inline comment explains the WHY well for anyone who hits this later.

PR #956, opened 2026-06-21
fix(site): prevent SSRF in Google Sheets submission

### LINKED ISSUE #935: SSRF in submitToGoogleSheets — URL is a client-controllable argument
## Issue

`submitToGoogleSheets` is exported as a server action and accepts a URL as a client-controllable argument, creating a Server-Side Request Forgery (SSRF) vector.

## Fix

Stop exporting the generic function. Only export env-bound wrappers that hardcode the target URL from environment variables.

BaoT1301:
## Description

Closes #935

`submitToGoogleSheets` was exported from a `'use server'` module while taking a caller-supplied URL. Because every export in a `'use server'` file becomes a directly-invokable server action, that meant any client could call it as an endpoint and make the server `fetch()` an arbitrary URL — a classic SSRF vector.

The fix is the one the issue recommends: stop exporting the generic URL-taking function and make it a private helper, so the destination can only come from a trusted, server-side source. The only exported entry point now is `submitContactToSheets`, which binds the target to `CONTACT_GOOGLE_SCRIPT_URL` on the server (and throws if it isn't configured), so a caller can't control where the request goes. I updated the one live caller (the contact form's `submitPublicInquiry`) to use the wrapper and dropped its now-redundant URL handling.

`apply` only referenced the function in a vestigial test mock (its action code doesn't import googleSheets), so nothing else needed touching.

### Testing
- Updated the contact action tests to the wrapper (forwards form data, never a URL).
- Added `googleSheets.test.ts` covering the security-relevant behavior: throws when `CONTACT_GOOGLE_SCRIPT_URL` is unset (fails closed) and posts only to the server-configured URL.
- Site tests for both files pass (7), lint clean.

## Checklist

### DIFF
--- a/apps/site/src/app/(unauthenticated)/[locale]/(contact)/contact/actions.test.ts
+++ b/apps/site/src/app/(unauthenticated)/[locale]/(contact)/contact/actions.test.ts
@@ -3,12 +3,12 @@
 import { beforeEach, describe, expect, it, vi } from 'vitest';
 
 const mocks = vi.hoisted(() => ({
-  submitToGoogleSheets: vi.fn(),
+  submitContactToSheets: vi.fn(),
   loggerError: vi.fn(),
 }));
 
 vi.mock('@/lib/actions/googleSheets', () => ({
-  submitToGoogleSheets: mocks.submitToGoogleSheets,
+  submitContactToSheets: mocks.submitContactToSheets,
 }));
 
 vi.mock('@/lib/logger', () => ({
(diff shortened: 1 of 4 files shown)


C0derTang (approved):
- Contact action tests correctly drop URL assertions + the now-dead missing-URL case (moved to wrapper).
- New googleSheets.test.ts covers both security behaviors: throws when env unset, posts only to configured URL. Env save/restore via ORIGINAL_URL is correct.

Looks good to me

PR #958, opened 2026-06-23
fix(site): remove internal routes from robots.txt (#938)

### LINKED ISSUE #938: Internal routes exposed in robots.txt
## Issue

`/admin/`, `/incoming/`, `/api/`, and other internal paths are listed in `robots.txt`, advertising attack surface to anyone who reads it. robots.txt is not a security control.

## Fix

Remove all internal/admin paths from `robots.txt`. Security through obscurity isn't a fix, but there's no reason to enumerate sensitive routes publicly.

rcjasub:
## Description

Fixes #938

Removes internal/admin paths (`/admin/`, `/api/`, `/_next/`, `/cdn-cgi/`, `/en-us*data=route*`, `/apple-app-site-association`) from `robots.txt` across all user-agent blocks. These paths were advertising internal infrastructure to anyone who reads the file. `robots.txt` is not a security control — real protection is handled at the application/middleware level — but there's no reason to enumerate sensitive routes publicly.

## Checklist

### DIFF
--- a/apps/site/public/robots.txt
+++ b/apps/site/public/robots.txt
@@ -6,39 +6,24 @@
 User-agent: *
 Crawl-delay: 1
 Allow: /
-Disallow: /signup/
-Disallow: /*/careers/externship
-Disallow: /login/
 Disallow: /*/*?*ls=*&ls=*
 Disallow: /*/*?*ls%3D*%3Fls%3D*
 Disallow: /*/*?*ls%3d*%3fls%3d*
-Disallow: /invite
-Disallow: /creator
-Disallow: /partner
+Disallow: /en-us*data=route*
 
 User-agent: AhrefsBot
 Crawl-delay: 10
-Disallow: /signup/
-Disallow: /*/careers/externship
-Disallow: /login/
 Disallow: /*/*?*ls=*&ls=*
 Disallow: /*/*?*ls%3D*%3Fls%3D*
 Disallow: /*/*?*ls%3d*%3fls%3d*
-Disallow: /invite
-Disallow: /creator
-Disallow: /partner
+Disallow: /en-us*data=route*
 
 User-agent: AhrefsSiteAudit
 Crawl-delay: 10
-Disallow: /*/careers/externship
-Disallow: /signup/
-Disallow: /login/
 Disallow: /*/*?*ls=*&ls=*
 Disallow: /*/*?*ls%3D*%3Fls%3D*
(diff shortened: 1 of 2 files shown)


C0derTang:
Pushed a follow-up commit (878e746) onto this branch to round out the fix, so we land #938 as one PR instead of two (I'd opened #959 in parallel — now closing it as superseded):

- **Removed the remaining internal routes** that were still enumerated in every crawler block: `/login/`, `/signup/`, `/invite`, `/creator`, `/partner`, `/*/careers/externship`. These are the same class of auth/internal paths #938 targets.
- **Restored `Disallow: /en-us*data=route*`** — it's a query-param crawl-hygiene rule (like the `?ls=` ones), not an internal/admin route, so it's out of scope for this issue and worth keeping to stop crawlers indexing RSC payloads.
- **Added `apps/site/src/app/robots.test.ts`** — a regression guard so internal routes can't silently creep back into the file.

### About the test

It reads the **real `public/robots.txt` from disk** (`readFileSync`, no mock/fixture), so it asserts against exactly what ships. Pinned to the node environment via `// @vitest-environment node` so the filesystem read is unambiguous regardless of the project's default (jsdom) test env. Six cases:

1. **No internal/admin routes leak** — loops `['/admin/', '/api/', '/login/', '/signup/', '/invite', '/creator', '/partner', '/cdn-cgi/', '/_next/', '/apple-app-site-association', 'careers/externship']` and asserts each is absent. This is the actual #938 guarantee, and it fails loudly if any future edit re-adds one.
2. **RSC data-route rule kept** — asserts `Disallow: /en-us*data=route*` is still present, so the crawl-hygiene rule isn't dropped again by mistake.
3. **Bad-bot blocks intact** — `AhrefsBot`, `CCBot`, `ByteSpider`, `Baiduspider` all present, plus a `Disallow: /` somewhere (the full-site bad-bot block survived).
4. **Anti-scrape disallows intact** — the `?ls=` param rules remain.
5. **Sitemaps intact** — both `sitemap/main.xml` and `sitemap/careers.xml` still declared.
6. **Site stays crawlable** — `User-agent: *` retains `Allow: /` (we removed disallows, we didn't accidentally block the whole site).

`bun run test robots` → **6 passing**.

rcjasub (commented):
Thanks for the follow-up commit — appreciate you rounding it out directly. Tests are passing, robots.txt looks clean. Ready to merge from my end.

C0derTang (approved):
LGTM

PR #965, opened 2026-06-26
fix: remove /disclosures page and update footer links to /legal

CaiZhengTech:
Removes the duplicate /disclosures route (redundant with /legal) and updates both site footers to link to /legal with label 'Legal'.

## Summary
- Removed duplicate `/disclosures` route (redundant with existing `/legal` page)
- Updated both site footers (marketing and Go/creators) to link to `/legal` with label "Legal"
- Removed disclosures translation files and cleaned up `message-files.ts`

## Test plan
- [ ] Footer "Legal" link navigates to `/legal`
- [ ] `/disclosures` returns 404
- [ ] All 474 unit tests pass (`bun validate` from `apps/site`)


rcjasub (commented):
Good cleanup overall — deletion is clean, footer tests kept in sync. Two things need to be addressed before this can merge:

Missing redirect (blocks merge)
/disclosures was a live public URL. Without a 301 redirect, any bookmark, backlink, or search-indexed URL will now hard 404. Please add to next.config.ts:
{ source: '/disclosures', destination: '/legal', permanent: true },
{ source: '/es/disclosures', destination: '/es/legal', permanent: true },

apps/site/.prettierignore — please remove
This file is an exact duplicate of the root .prettierignore (same 12 lines). The root already covers apps/site/ via **/ patterns so this adds no behavior. It's redundant and should be deleted rather than kept.

.husky/commit-msg — out of scope, needs its own PR
The npx → bunx swap is a real fix (known Windows + Husky issue I think based on what I experience) but it doesn't belong in this PR. Please revert it here and open a dedicated PR so the fix gets proper visibility and context.

Dropping this —  Nothing left to link to.
Content gap on /legal (blocks merge)
The disclosures page linked to name_brand_identity_misues.pdf and modern_slavery_act_statement.pdf. Checked /legal — it's a plain text list with no links at all. Those PDFs will have no navigation entry point after this merges. Either add links to them on /legal, or confirm they're intentionally being removed from the site.

vincenttodd:
the pdf were intentionally removed

vincenttodd:
shut up copilot

rcjasub (approved):
LGTM

PR #967, opened 2026-06-27
fix(security): patch linkify-it high-severity advisory via override

rcjasub:
## Description

Fixes GHSA-22p9-wv53-3rq4 — patches `linkify-it <=5.0.0` high-severity vulnerability (quadratic scan loop complexity) by adding a `^5.0.1` override in the root `package.json`. `linkify-it` is a transitive dependency via `markdown-it` → `sanity`, no direct usage in the codebase.

## Checklist

- [x] You've used conventional commits where appropriate.

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -224,6 +224,7 @@
     "flatted": "^3.4.2",
     "form-data": "^4.0.6",
     "js-yaml": "^4.1.1",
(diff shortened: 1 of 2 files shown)


C0derTang (approved):
dependency update doesnt seem to break anything, lgtm

Issue #934, opened 2026-06-12
CSP allows unsafe-eval and unsafe-inline, defeating XSS protection


PratGit1606:
## Issue

The Content Security Policy includes `unsafe-eval` and `unsafe-inline`, which completely defeats XSS protection. With Stripe active, a successful injection could overlay a fake payment form.

## Fix

Switch to nonces for inline scripts. Both PostHog and Stripe support nonce-based CSP. Remove `unsafe-eval` and `unsafe-inline` from the policy.

Orlando275:
Partial update on where this stands, with the measurements in https://github.com/toddagriscience/nightcrawler/issues/261#issuecomment-5414904886.

**The `unsafe-eval` half is already done.** `buildScriptSrc` in `next.config.ts` includes it only in development, and `next.config.test.ts` guards that it never reaches the production policy.

**The `unsafe-inline` half cannot be delivered with nonces on this codebase.** `cacheComponents: true` (PPR) is incompatible with nonce-based CSP, and a live production response carries 22–29 executable inline scripts — the React Flight payload — none of which can take a per-request nonce from a prerendered shell, and whose per-page, per-build contents rule out hashes too.

Worth noting the app contributes **no inline scripts of its own**: the only ones are Next.js's, plus two `application/ld+json` data blocks. So the remaining exposure is narrower than the title suggests.

Before tightening anything, the missing piece is visibility: there is no `report-uri` or `report-to` in the policy today, so no one can see what a strict policy would actually break. A `Content-Security-Policy-Report-Only` header alongside the enforced one would collect that at zero risk.

PR #968, opened 2026-06-28
fix(site): drop unsafe-eval from production csp

BaoT1301:
## Description

Part of #934 (does not fully close it — see Scope below).

Removes `'unsafe-eval'` from the site's Content-Security-Policy `script-src` directive in production, keeping it only under `next dev`.

**What:** `script-src` is now built by a small exported helper, `buildScriptSrc(dev)`, instead of a hardcoded string:

```ts
script-src 'self' 'unsafe-inline' https://*.posthog.com https://js.stripe.com
// + 'unsafe-eval' only when NODE_ENV === 'development'
```

**Why:** The CSP allowed `'unsafe-eval'` in every environment. `'unsafe-eval'` permits executing strings as code via `eval` / `new Function`, which largely defeats the XSS protection CSP exists to provide — and with Stripe live, a successful injection is high-impact. Per Next.js's CSP guidance, `'unsafe-eval'` is only needed in development (React uses `eval` to reconstruct server-side error stacks); neither React nor Next.js use `eval` in production by default, and PostHog (`posthog-js`) and Stripe.js don't require it either, so it's safe to drop there.

**How:**
- Added `buildScriptSrc(dev)` in `apps/site/next.config.ts`: conditionally includes `'unsafe-eval'` based on `NODE_ENV === 'development'`.
- Wired it into the `Content-Security-Policy` header.
- Tests in `apps/site/src/next.config.test.ts`: `buildScriptSrc(true)` includes `'unsafe-eval'`, `buildScriptSrc(false)` does not, both keep `'self'`, `'unsafe-inline'`, PostHog and Stripe; plus a regression test on the real config asserting the production policy contains no `'unsafe-eval'`.

### Scope / follow-up

This is a scoped step, not the full nonce migration. #934/#261 also ask to remove `'unsafe-inline'` via nonces, but per-request nonces (and the `'strict-dynamic'` they require) conflict with this app's `cacheComponents` (PPR) rendering — the static shell's framework scripts have no nonce. That's the same wall that got the earlier attempt (#258) discarded. Dropping `'unsafe-eval'` is the high-value, low-risk part that works without forcing every page to dynamic rendering or risking the live Stripe checkout. Left a note on #934 outlining the nonce blocker for a maintainer to weigh in on.

## Checklist


henryng15 (approved):
Looks good to me

PR #969, opened 2026-06-28
feat(site): add help link to account side menu

### LINKED ISSUE #734: Add a help link to platform account pages
### Problem Statement

Theres no easy way to contact us on the account page. We could either add a help link to the page selector on the left or in the top right of the page.

### Context

_No response_

### Constraints

_No response_

### Technical Details

_No response_

### User Story

_No response_

### Use Case

_No response_

BaoT1301:
## Description

Fixes #734

There was no easy way to reach support from the account pages. This adds a **"Help"** link to the account side menu (the left-hand page selector), in the same bottom utility group as "Log out", pointing to the existing authenticated `/contact` ("Support Tools") page — so users land on the email/contact options already built there.

**What changed**
- `account-side-menu.tsx`: new "Help" link to `/contact`, with a `BiHelpCircle` icon trailing the label (mirroring the "Log out" treatment) and the label aligned flush-left with the section nav items.
- A unit test asserting the Help link renders and targets `/contact`.
- A Storybook story (`Default` and `PrivacyActive`) so the menu can be viewed in isolation.

**Notes / decisions**
- Reuses the existing `/contact` Support Tools page rather than introducing a new help destination — smallest, most consistent fix.
- The issue suggested the left page selector *or* the top-right; I went with the left side menu as the lower-risk, cleaner spot. Easy to move to the header if preferred.
- Follows this component's existing conventions: plain text label (the authenticated account area is not under `[locale]`, so it is not internationalized) and the same typography/hover styling as the section nav links.

## Checklist

### DIFF
--- a/apps/site/src/app/(authenticated)/(shell)/(accounts)/account/components/account-side-menu/account-side-menu.stories.tsx
+++ b/apps/site/src/app/(authenticated)/(shell)/(accounts)/account/components/account-side-menu/account-side-menu.stories.tsx
@@ -1,0 +1,40 @@
+// Copyright © Todd Agriscience, Inc. All rights reserved.
+
+import type { Meta, StoryObj } from '@storybook/nextjs-vite';
+import AccountSideMenu from './account-side-menu';
+
+const meta: Meta<typeof AccountSideMenu> = {
+  title: 'Account/AccountSideMenu',
+  component: AccountSideMenu,
+  parameters: {
+    layout: 'padded',
+    nextjs: {
+      appDirectory: true,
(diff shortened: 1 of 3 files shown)


rcjasub (approved):
Nice addition! One small cleanup worth considering: the <span className="text-sm font-normal"> inside the Help link (line 61) duplicates classes already set on the parent <Link> — those properties cascade down so the span's className does nothing. The same pattern exists on the Log out button's inner span (line 70). Neither is a blocker, just dead markup that could be simplified to plain text nodes

PR #973, opened 2026-06-30
fix(account): correct principal operator name key and guard against m…

### LINKED ISSUE #971: [Bug]: db/loading issue in Vercel test environment for /account pages
### Bug Description

Clicking on the /account link in the new left menu yields the basic error page. Could be an issue with the db. 

### Steps to Reproduce

on the home view in Iris, click on Account link in left menu

### Expected Behavior

should load the /account page

### Screenshots/Videos

_No response_

### Browser

Safari

### Device Type

Desktop

### Language/Locale

None

### Additional Context

_No response_

rcjasub:
Title: fix(account): correct principal operator name key and guard against missing farm record (#971)

Summary

- Fixed a key mismatch in getAccountUsersData() where principalContact was returning name but users/page.tsx was reading firstName, causing the Principal Operator name to always render blank on /account/users
- Added a null guard in getAccountFarmData() to throw a descriptive error if no farm record is found, preventing a silent TypeError crash on /account

Changes

apps/site/src/app/(authenticated)/(shell)/(accounts)/account/db.ts
- Renamed name → firstName in principalContact to match what users/page.tsx expects and to be consistent with how the owner object is shaped
- Added if (!farmRecord) check before accessing .farm and .farm_location on the destructured query result

Root Cause

Both bugs were isolated to db.ts:

1. The principalContact object was built with a name key but the page consumed it as firstName. TypeScript didn't catch this because the return type is Partial<UserSelect>, which allows all fields to be undefined — so accessing a missing key silently returned undefined instead of throwing.
2. Drizzle's destructured query result (const [farmRecord] = await db...) returns undefined if no rows match. Without a guard, accessing farmRecord.farm would throw a TypeError if a user's farm record is missing from the DB.

Testing

- Navigate to /account/users — Principal Operator name should now display correctly
- Verify /account loads without errors for authenticated users

Closes #971


Jose-Gael-Cruz-Lopez:
## Still relevant — both bugs are live on `main` today. This is a cheap, high-value merge.

The conflict here is **purely positional**, not semantic: the file moved. Worth rescuing rather than letting age bury it.

### Bug 1 is user-visible in production right now

`apps/site/src/app/(authenticated)/(no-sidebar)/account/db.ts:69-76` builds:

```ts
const principalContact = { name: toDisplayName(...), email: ..., phone: ... };
```

and returns it as `principalOperator`, typed `Partial<UserSelect>`. But the consumer at `…/account/(with-shell)/users/page.tsx:17` reads:

```tsx
value={principalOperator.firstName}   // undefined
```

So the **Name row under "Principal operator" renders empty**. The fix in this PR (`name` → `firstName`) is confirmed correct by the sibling branch three lines below — `db.ts:82` already uses `firstName: toDisplayName(...)` for `owner`.

Why TypeScript does not catch it: excess-property checking only applies to *direct* object-literal assignment. Because the literal is bound to the intermediate `principalContact` variable first, and `Partial<UserSelect>` makes every key optional, `{ name: … }` passes `type-check` silently.

### Bug 2 is an unhandled 500

`db.ts:96-103`:

```ts
const [farmRecord] = await db.select()....limit(1);
return { farm: farmRecord.farm, location: farmRecord.farm_location };
```

No guard. A missing farm row is a `TypeError: Cannot read properties of undefined` → unhandled 500 on the account page. Note `getAccountShellData` at `db.ts:34-35` already uses optional chaining, so the guard restores intra-file consistency.

### The conflict

The PR targets `apps/site/src/app/(authenticated)/(shell)/(accounts)/account/db.ts`. On `main` that file is `apps/site/src/app/(authenticated)/(no-sidebar)/account/db.ts` — the `(shell)/(accounts)` route-group pair was collapsed to `(no-sidebar)` in #1005, and the page moved under `(with-shell)/users/`. Content also drifted: the PR's hunks target lines 65-71 and 98-104; the equivalents on `main` are 69-76 and 96-103.

Reapplying the 5 lines at the new path should take a couple of minutes.

### Please also fix the test — it is why this survived 52 days

`…/(with-shell)/users/page.test.tsx:10` mocks the key as **`fristName`** — a typo, so neither `name` nor `firstName`. And the assertions at `:26-28` only check `alex@example.com`, `Jamie Admin` and `jamie@example.com`; the principal operator's *name* is never asserted. The test therefore passes both before and after this change.

Suggested:

```diff
-      fristName: 'Alex',
+      firstName: 'Alex',
```

plus `expect(screen.getByText('Alex Owner')).toBeInTheDocument();` — otherwise this ships unverified and can silently regress.

### One style note

`throw new Error('Farm not found')` surfaces the bare message through `app/error.tsx`, which renders `{error.message}` at `:24`. Comparable "not found" paths in this repo use `notFound()` from `next/navigation` (see `contact/page.tsx:20`) or `throwActionError` (see `apply/actions.ts:261`). Since this is a page data loader, `notFound()` is probably the right semantic.

rcjasub (commented):
Fixed in the latest push:
- `name` → `firstName` in `principalContact` (db.ts) — Principal Operator name now renders
- Missing farm record now calls `notFound()` instead of throwing a bare Error
- Test's `fristName` typo fixed, added an assertion for the principal operator's name so this regression would actually be caught

Branch was also brought up to date with main (was based on a pre-#1005 commit, before the account routes moved to (no-sidebar)).

Jose-Gael-Cruz-Lopez:
Closing: superseded by #1083, which merged. Same bug — `getAccountUsersData` returned `name` while the page read `.firstName`, so the operator's Name row rendered blank. #1083 fixed it the other way round: it keeps `name` in the data layer (it holds `First Last`, so `firstName` would be a misnomer) and updates the page to read `.name`, plus adds an `AccountContact` interface and db-layer tests. Thanks for spotting it.

PR #975, opened 2026-06-30
feat(site): zone template page with 4 cards, remove widget grid

longlong9889:
## Description

Problem statement   

The authenticated home page (/) rendered each management zone as a draggable widget grid. Widgets were never the long-term direction for the zone view — the plan is a structured, advisor-driven layout where each management zone shows its soil-analysis info, mineral-based recommendations (IMPs), client observations, and search in a consistent template.

Context

This is the first step toward the zone-based dashboard. The full vision (per design) is a per-zone template with four cards. The data layer for the mineral IMPs (the analysis / mineral / oxidationRate / solubility tables and the recommendation-matching logic) isn't wired up yet and depends on an in-progress source-of-truth decision with the agronomy side, so the mineral IMP card ships as a placeholder for now.

What this PR does

  - Replaces the widget grid on / with a 4-card zone template (ZoneTemplate), rendered for the currently selected management zone:
    a. Info — zone name, sample date (from the latest soil analysis), next sample date (sample + 6 months), and lead advisor (placeholder — no advisor field exists yet).
    b. Mineral IMPs — "Info not currently available" placeholder until the analysis tables are wired.
    c. Observations — card shell with a (currently non-functional) "Add an observation" affordance; the backing table is a follow-up.
    d. Search — placeholder search input.
  - Stops rendering the widget grid (CurrentTab / WidgetsGrid / AddWidgetDropdown) on the zone page. Zone selection via ?zone=<id> and the sidebar are unchanged.
  
Testing

  - Verified locally: the 4 cards render for the selected zone, dates compute correctly from a real analysis (Sample → Next = +6 months), zone switching via the sidebar updates the cards, and the empty state still shows for farms with no zones.

### DIFF
--- a/apps/site/src/app/(authenticated)/(shell)/(widgets)/components/zone-template.tsx
+++ b/apps/site/src/app/(authenticated)/(shell)/(widgets)/components/zone-template.tsx
@@ -10,7 +10,12 @@ const TEST_PERIOD_MONTHS = 6;
 /** Placeholder until advisors are modeled in the DB. */
 const LEAD_ADVISOR_PLACEHOLDER = 'Not assigned';
 
-const CARD_CLASS = 'rounded-xl border border-stone-200 p-6';
+/** Shared vertical spacing for each stacked section. */
+const SECTION_CLASS = 'py-6';
+
+/** Small uppercase, letter-spaced section label. */
+const LABEL_CLASS =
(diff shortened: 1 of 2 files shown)


vincenttodd (approved):
Looks great!

PR #979, opened 2026-07-01
feat(site): inference search popup, panel styling, and iris nav icon

vincenttodd:
Add command-palette search that opens a semantic results panel with chat-style questions and summary blocks. Restore mineral IMP placeholder copy, update sidebar navigation, and replace the contact icon with the iris mark.

## Description

Fixes #{issue number}

## Checklist


Jose-Gael-Cruz-Lopez (approved):
## Approving ✅

Reviewed the full diff plus the five follow-up `fix:` commits, and verified the search flow end-to-end against a local Postgres 18 + pgvector build (migrations, seed, login, and a real `soil pH` inference search returning IMP + seed results).

### Verified
- **Iris icon** replaces the sparkle on the command-palette suggestion row.
- **Mineral IMPs** heading removed; "Information temporarily not available" retained.
- Results panel header stripped of the search icon / "Search" label / divider; **follow-up bar** is now a bordered box with a `BiUpArrowAlt` submit button; question bubble uses a uniform corner radius.
- Collapse control uses `BiDockLeft`; the edge tab is gone, replaced by a `BiDockRight` affordance in the top-right; sidebar toggle now uses the dock icons.
- CI is green (Build, Unit Tests, Code Quality, Security Audit). Confirmed locally that the suite's failure profile is identical with and without these commits, so they introduce no test regressions (the local failures are a pre-existing `localStorage` env quirk that does not affect CI).

### Non-blocking observations (follow-ups, not required for merge)
- `⌘`+number zone shortcuts (`zone-keyboard-nav`) collide with browser tab-switching in most browsers, so the advertised `1 ⌘` badges may not fire; consider a non-conflicting chord.
- `actions/search-modal.ts` (+ its test) appear to be orphaned now that the browse modal is gone; safe to delete in a follow-up.
- The command-palette popup sets `aria-modal` without a focus trap.

Nice work — the inference search feels good.

PR #980, opened 2026-07-01
feat(db,site): general IMP mirror + search over IMPs and varieties

longlong9889:
Add general_imp table + re-runnable sheet importer (33 IMPs) modeled on the
  variety importer. Wire the search panel to browse/filter both IMP sources (general + plan) and all seed varieties (link to /v).

TODO: lazy-load the ~943 varieties in the
  search panel instead of loading them all on open.

## Description

Fixes #{issue number}

## Checklist


C0derTang (dismissed):
## Must fix before merge

**1. Re-importing IMPs can overwrite the wrong row**
`packages/db/scripts/import-general-imps.ts`
When two IMPs make the same slug (for example two rows both named "Copper"), the first one gets the plain slug `copper` and the second gets a slug with a hash on the end. The problem is that "first one" depends on the row order in the sheet. If someone reorders the sheet, the second Copper can now grab `copper`, match the *other* Copper's database row, and overwrite its title, tags, content, and embedding. The original row gets a new id and the old one is deleted. So a harmless reorder can silently swap two practices and change their ids. Anything pointing at a `general_imp.id` would now point at the wrong practice. The identity needs to be stable no matter the row order.

**2. The prod migration workflow can report success while leaving prod half-migrated**
`.github/workflows/prod-db-reconcile.yml`
It decides whether a migration is "already applied" by checking a single object (for example, for migration 0012 it only checks the `seed_variety` table). But some migrations create several things. 0012 also creates `seed_crop` and three foreign keys, and 0011 also renames an enum and adds a column. If only the one probed object exists (say, from a partial manual fix), the workflow skips that migration and then writes it into the migration ledger as "done." After that, nothing will ever repair the missing pieces, and the run shows green. Related: the 0011 rename guard only refuses when *both* the old and new tables are missing; it should also refuse when *both* exist. And the ledger-rebuild step exits successfully (just a warning) when the schema isn't actually complete, so a broken run still looks like a passing run.

**3. A real search outage looks like "no results"**
`apps/site/src/lib/ai/search.ts`
We wrap each search source so one broken source doesn't take down the whole search, which is good. But all three sources share the same embedding vector and database join. If something systemic breaks (for example the embedding size stops matching the database column), all three fail at once, each quietly returns nothing, and the user sees "No matching results, try rephrasing" instead of an error. A total outage is indistinguishable from a normal empty search. If *every* source fails, we should treat that as an error, not an empty result.

## Should fix

**4. Search box wildcards match everything**
`apps/site/src/app/(authenticated)/actions/search-imps.ts`
The keyword search drops what the user typed straight into a SQL `ILIKE` pattern. It's safe from injection (the value is parameterized), but `%` and `_` are wildcards. So typing just `%` matches every IMP, and something like `50_off` treats `_` as "any character" and over-matches. Those characters should be escaped before building the pattern.

**5. Fast typing can show stale search results**
`apps/site/src/app/(authenticated)/components/search-panel/search-popup.tsx`
The command palette fires a search as you type, but nothing guards against responses coming back out of order. Type "cop", then "per" quickly; if the "cop" response arrives after the "copper" one, the older results overwrite the newer ones and the list no longer matches the box. It needs a request-id (or abort) check so only the latest response wins.

**6. A stuck request spins "Searching..." forever**
`apps/site/src/app/(authenticated)/components/search-panel/search-panel-context.tsx`
If the search request hangs (slow network, dropped connection), neither the success nor the error path ever runs, so the panel stays on "Searching..." with no way out except resubmitting. There's also no cleanup, so if the user navigates away mid-search it tries to update state on a screen that's gone. A timeout or abort would fix both.

**7. Prod database connection doesn't check the server name**
`.github/workflows/prod-db-reconcile.yml`
The connection uses `sslmode=verify-ca`, which checks the certificate is signed by the trusted authority but does **not** check it's actually the right server. For a job sending the production DB password, `verify-full` is safer. If the provider's certificate genuinely can't match the hostname, a comment saying so would help.

## Nice to clean up

**8. Two search paths that must be kept in sync by hand**
There are now two ways to search IMPs, the keyword one (`search-imps.ts`) and the semantic one (`search.ts`), each with its own result shape and its own copy of "what title to show when there's none" and "how to trim the preview." They already disagree: an untitled IMP shows "Management practice" in the palette but "Integrated Management Practice" in the semantic panel, and the fallback even reads a different tag position (`tags[0]` vs `tags[1]`). The link to the IMP page is also hardcoded in the popup while the panel has a `resultHref()` helper for the same thing, so renaming the route silently breaks the popup. Worth pulling the shared bits (title fallback, preview, routing) into one place.

**9. Some search fields are computed but never shown**
`apps/site/src/lib/ai/search.ts`
`category` and `source` are filled in for every result but nothing displays them, and the seed query does an extra table join purely to build `source`. Either use them or drop them (and the join).

## Small stuff
- Dry-running the importers no longer checks you're pointed at localhost (the check moved into the commit-only path), so a misconfigured target isn't caught until you add `--commit`.
- The CSV parser keeps `\r` characters inside quoted multi-line cells, so a sheet exported with Windows line endings can trigger needless re-embedding.
- The ledger rebuild uses `TRUNCATE` without `RESTART IDENTITY`, so the ids won't be 1 to 15 as the surrounding comment implies (cosmetic only).

C0derTang:
Addressed all review findings in 870943c:

**Must fix**
1. **Slug order-dependence** (`import-general-imps.ts`) — slug assignment is now two-pass: base slugs are counted across the whole sheet first. A base slug that occurs once is used bare; when duplicated, *every* duplicate gets a content-hash suffix (no first-row winner), so reordering rows can no longer swap DB identities. The misleading comment was rewritten to describe the new rule.
2. **Prod reconcile workflow** — the verify step now probes every object each migration creates (tables via `to_regclass`, enums via `pg_type`, columns, FK constraints). Three states per migration: all present → skip, none present → apply, partial → `::error` naming present/missing objects + exit 1, so a half-applied migration can never be skipped and written into the ledger. The 0011 guard also refuses when *both* `platform_access_application` and `form_submissions` exist. The ledger rebuild now errors (exit 1) instead of warning when the schema isn't fully at 0014.
3. **Search outage vs. empty results** (`search.ts`) — `runSource` now returns `null` on failure instead of `[]`. If all three sources fail, `searchKnowledge` throws (`All search sources failed`) and the UI shows the error state; partial failures still degrade gracefully.

**Should fix**

4. **ILIKE wildcards** (`search-imps.ts`) — `\`, `%`, `_` are escaped before building the pattern, so `%` no longer matches everything.
5. **Stale popup results** (`search-popup.tsx`) — added a request-id guard (same pattern as the panel context); only the latest response applies, and clearing the query invalidates in-flight requests.
6. **Stuck "Searching…"** (`search-panel-context.tsx`) — 30s timeout flips to the error state and clears `isSearching`; effect cleanup invalidates the in-flight request id so an unmounted tree never receives state updates.
7. **TLS** — all three psql connections use `sslmode=verify-full` (hostname validation), with a comment noting a downgrade back to `verify-ca` needs explicit justification.

**Cleanup**

8. **Shared display logic** — new `search-panel/search-display.ts` exports `impDisplayTitle` (single fallback string, "Integrated Management Practice", matching the detail page), `resultHref` (moved out of `search-panel-body.tsx`; the popup uses it instead of a hardcoded path), and `previewText` (single trim/ellipsis implementation for both surfaces).
9. **Dead fields** — `category`/`source` removed from all three source queries, the result mapping, and `SearchResult`; the seed query's `leftJoin` that existed only to build `source` is gone.

**Small stuff**
- Both importers validate the DB target up front, so a misconfigured target is caught on dry runs too.
- `parseCsv` normalizes `\r\n`/`\r` to `\n` before parsing, so Windows line endings inside quoted cells can't change content hashes and force re-embeds.
- Ledger rebuild uses `TRUNCATE … RESTART IDENTITY` so ids are 1–15 as the comment implies.

Verified: `tsc --noEmit` clean in `apps/site` and `packages/db`, lint clean (2 pre-existing warnings in unrelated marketing tests), 125/125 vitest tests pass in the authenticated area, workflow YAML parses and embedded bash passes `bash -n` with the partial-state classifier smoke-tested.

C0derTang (approved):
see above comments

PR #1001, opened 2026-07-02
feat: add soil test and content parsers to monorepo

henryng15:
## Description

Cherry-picks the soil test parser (Parser 1) and content-to-markdown parser (Parser 2) from the stale `feat/soil-test-parser` branch onto a fresh branch from main. The original branch was 103 commits behind and used the old `frontend/` directory structure — this port maps everything to the current monorepo layout.

## What's included

**`packages/db` schema changes**
- `mineral_types` enum: added `Manganese`, `Copper`, `Boron`
- `units` enum: added `dimensionless`
- New `mineral_tag` enum (`Low`, `Med`, `High`)
- New `mineral` table columns: `idealValue`, `tag`, `fourLows`
- `get-standard-values.ts`: null-maps the three new minerals (no standard_values columns for them yet)

**`apps/site/scripts/` — new developer scripts**
- `parse-soil-test.ts` — reads CSV/Excel lab reports and inserts into `analysis`, `mineral`, `solubility` tables
- `parse-content.ts` — CLI wrapper for the content parser; supports `--save` to embed content into the knowledge base
- `migrate-mineral-schema.ts` — one-time migration to bring an existing local DB up to the new mineral schema
- `cleanup-first-run.ts` — removes first-run test data from farm_id=2

**`apps/site/src/lib/ai/content-parser.ts`** — core parser library (images via vision AI, PDFs via text extraction, Excel to markdown)

## Notes

- No migrations generated (backend team handles that per project rules)
- `pdf-parse`, `exceljs`, and `openai` added as deps to `apps/site/package.json`
- The `--save` path in `parse-content.ts` stores only the embedding; `knowledgeArticle` on main no longer has title/content/category columns

## Checklist


Jose-Gael-Cruz-Lopez:
Hey, one domain question on the pH logic before this merges.

In `parse-soil-test.ts`, `IDEAL_VALUES.PH` is set to `6.75` (the midpoint of the 6.5 to 7.0 ideal range), but `computeTag` for PH tags anything below 7.3 as `Low`. So a reading sitting right at the ideal (6.75) would come back tagged `Low`, which seems off.

The tag thresholds (7.3 / 7.8) look more like an alkalinity scale than the 6.5 to 7.0 ideal range, so it feels like the tag and the ideal are using two different frameworks. Can you confirm whether that's intended, or should the tag thresholds line up with the ideal range?

Everything else looks good on my end.

henryng15:
Good catch - intentional. `idealValue` comes from Vincent's Data Framework (ideal 6.5–7.0, so 6.75), `tag` comes from a separate lab reference range (Low <7.3, Med 7.3–7.8). Two different scales by design - same split applies to every mineral (e.g Potassium's ideal is 1200 PPM but tags High since lab-range medium is only 250–350).

> Hey, one domain question on the pH logic before this merges.
> 
> In `parse-soil-test.ts`, `IDEAL_VALUES.PH` is set to `6.75` (the midpoint of the 6.5 to 7.0 ideal range), but `computeTag` for PH tags anything below 7.3 as `Low`. So a reading sitting right at the ideal (6.75) would come back tagged `Low`, which seems off.
> 
> The tag thresholds (7.3 / 7.8) look more like an alkalinity scale than the 6.5 to 7.0 ideal range, so it feels like the tag and the ideal are using two different frameworks. Can you confirm whether that's intended, or should the tag thresholds line up with the ideal range?
> 
> Everything else looks good on my end.

vincenttodd:
@longlong9889 sorry to throw one more thing on your plate, I just want you to review this before its ready to merge

longlong9889 (changes requested):
hey Henry, this PR changes the mineral schema (`mineral.ts`: the new `Manganese`/`Copper`/`Boron` enum values, `dimensionless` unit, the `mineral_tag` enum, and the`ideal_value`/`tag`/`four_lows` columns) but doesn't include a Drizzle migration. Per our workflow that's expected, contributors don't run `drizzle-kit generate`, the migration is handled separately.

2 issues we need to solve before merging:

- The columns won't reach staging/prod without a committed migration. Since the fail-loud migration system landed, CI only applies committed `.sql` files, it  doesn't generate them. With no `0015`, `getMineralsForAnalysis` had to be patched to 
a hardcoded column list to avoid "column does not exist," and those columns never get created downstream.
- Also, this branch is behind main (main is at `0014_general-imp`, this branch at `0013`), so any migration has to be generated *after* a rebase or it'll collide with `0014`.

I've generated it locally and created a separate PR for this

We can either:

  - Keep this PR as-is: rebase your branch onto main, then cherry-pick my `0015`
    commit from `feat/mineral-migration` onto it. This PR stays the canonical one.
  - Or i merge `feat/mineral-migration` instead cuz it already has your work rebased on main + the migration together. Simpler git-wise but it takes the feature off your branch. Either works for me. 

  Separately, a few review notes to address wherever this lands:
  - Scripts document `NODE_TLS_REJECT_UNAUTHORIZED=0`, which disables TLS verification process-wide — prefer wiring the CA cert like the importers do.
  - ExcelJS cells aren't always primitives (formulas/rich text/dates) — `String(cell)` can
    yield `"[object Object]"` and numeric formula cells get silently dropped.
  - `cleanup-first-run.ts` hard-deletes all `farm_id=2` data with no confirmation guard.
  - No tests on the conversion math (`buildMinerals`/`computeTag`/`checkFourLows`/meq→PPM).

henryng15:
### Update — rebased on main + migration + review notes

Went with option 1 from @longlong9889's review: rebased this branch onto `main` (was at `0013`, main is now past `0014`) and cherry-picked the `0015` migration commits, so this stays the canonical PR with the migration commit authored by longlong intact. The branch tree matches `feat/mineral-migration` exactly, plus the follow-up fixes below.

**Review notes addressed**

- **ExcelJS cell extraction** — added `excelCellToString` to unwrap formula / rich-text / hyperlink / error cells at the single read site, so they no longer stringify to `"[object Object]"` or silently drop formula results. Unit-tested.
- **Conversion math tests** — extracted the meq/l→PPM conversion, lab-range tags, and Four Lows detection out of `parse-soil-test.ts` into `scripts/lib/soil-conversion.ts` (no behaviour change) and covered `computeTag` / `buildMinerals` / `checkFourLows` with unit tests.
- **`cleanup-first-run.ts`** — now inspects and reports by default; deletes only when passed `--confirm`.

**Deferred**

- **TLS / CA cert wiring** — the scripts still document `NODE_TLS_REJECT_UNAUTHORIZED=0`, which the shared `connection.ts` requires to reach a local Docker DB (it has no plaintext path otherwise). Doing this properly means routing these scripts through the importer's verify-ca target resolver, which I can't validate without a live remote DB, so I've left it as a follow-up rather than ship an unverified connection change.

`bun validate` passes end to end (format → type-check → lint → test → build); CI is green.

longlong9889 (approved):

PR #1003, opened 2026-07-03
feat(db,ci): fail-loud remote migrations, env-aware reconcile + importers

C0derTang:
## Why

Prod silently drifted behind `main` because CI applied schema changes with `drizzle-kit push`, which aborts on its interactive enum resolver in a TTY-less runner while still exiting 0 — the "Push to production" step reported success while applying nothing (that's how `/v` broke on a missing `seed_variety` table). This PR removes that failure mode and finishes the plumbing needed to roll out `general_imp` (migration 0014) + its data to staging and prod.

## What

- **New `packages/db/scripts/migrate-remote-db.ts`** (`--env staging|prod`): applies committed migrations with the ledger-based drizzle-orm migrator (same mechanism as `migrate-local-db.ts`), then asserts the ledger covers every journal entry. Any failure exits non-zero.
- **CI workflows now migrate instead of push.** Both the main-push (staging) and release (prod) DB workflows run the migrator against the pg_dump clone first (a true dry-run — the clone carries the ledger) and then against the real database. The stale "push because drizzle-kit is scuffed" path is gone.
- **Reconcile workflow is environment-aware.** `prod-db-reconcile.yml` → `db-reconcile.yml` with a required `environment` input (`staging`/`prod`); adds a `knowledge_article` baseline probe (refuses to run against a wrong/empty DB) and builds the ledger rebuild rows at runtime from `meta/_journal.json` + `sha256sum` (verified byte-identical to the previously hardcoded 15 rows) so 0015+ never require editing the workflow.
- **Importers gain a guarded remote mode** via `scripts/lib/db-target.ts`: `--env staging|prod` for dry runs; committing remotely additionally requires `--confirm <env>` and `OPENAI_EMBEDDINGS_KEY` (placeholder embeddings must never reach a shared DB), plus a parsed-row floor so a truncated CSV export can't prune a live mirror. Default local behavior unchanged.
- **Fixes:** `STAGING_DATABASE_PEM_CERT` was read by `drizzle-staging.config.ts` but never set in CI (now falls back to `DATABASE_PEM_CERT`); documented the display-name coupling between the release workflow and `vercel-deploy-on-release.yml`.

## Verification

- `tsc --noEmit`, `eslint`, and the package test suites pass.
- Fresh local DB: migrator applies 0000–0014, ledger = 15 entries, `general_imp` exists; re-run is a no-op; bad target/credentials exit 1 loudly.
- Importer gates verified: missing `--confirm`, wrong `--env`, and missing `OPENAI_EMBEDDINGS_KEY` all refuse before touching anything; remote dry-run parses without connecting.
- Runtime-generated ledger rows match the old hardcoded values exactly.

## Rollout (after merge)

1. The main-push staging workflow will fail red once — expected, staging has no ledger yet.
2. Actions → **DB Reconcile**: `prod` `verify` → `apply`, then `staging` `verify` → `apply` (creates `general_imp` + rebuilds ledgers).
3. Re-run **Database Compatibility Check** — should be green end-to-end.
4. Cut a release → prod migrates (no-op) → Vercel deploy chains as before.
5. Load the 33 IMPs per env: `bunx tsx scripts/import-general-imps.ts --env staging` (dry run) then `--commit --confirm staging`; repeat for `prod`.

### DIFF
--- a/.agent/workflows/modifying-database.md
+++ b/.agent/workflows/modifying-database.md
@@ -4,3 +4,10 @@ description: When modifying database tables
 
 1. Apply changes to the schema.
 2. DO NOT create migrations or run migrations (e.g., absolutely no running `bunx drizzle-kit generate`). Let the specific backend teams handle migrations.
+
+## How migrations reach staging/prod (for the backend team)
+
+- Committed migrations live in `packages/db/drizzle/*.sql` + `meta/_journal.json`. CI applies them with the ledger-based drizzle migrator (`scripts/migrate-remote-db.ts`), never `drizzle-kit push` — push silently no-ops in CI (interactive enum resolver, no TTY, exits 0), which is how prod once drifted.
+- Merge to `main` → "Database Compatibility Check" workflow migrates **staging**. Cutting a release → "Release Database Compatibility Check And Push" migrates **prod**, then chains the Vercel deploy. Both fail loudly on any mismatch.
+- If a database has drifted (schema present but ledger disagrees), run the manual "DB Reconcile" workflow (Actions → workflow_dispatch → pick `staging`/`prod`, run `verify` first, then `apply`).
+- Row data for sheet-mirrored tables (seed varieties, general IMPs) is loaded by the importers in `packages/db/scripts/import-*.ts`; they default to the local Docker DB, and can target remotes with `--env staging|prod` plus `--commit --confirm <env>` and `OPENAI_EMBEDDINGS_KEY`.
--- a/.github/workflows/db-reconcile.yml
+++ b/.github/workflows/db-reconcile.yml
@@ -129,11 +129,24 @@ jobs:
           # this workflow handles. If it is missing we are pointed at a wrong
           # or empty database and must not proceed (an apply run would rebuild
           # a ledger full of claims about schema that does not exist).
-          baseline=$(has_table knowledge_article)
-          if [ "$baseline" != "t" ]; then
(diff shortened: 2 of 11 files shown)


rcjasub (commented):
Overall: solid fix for the CI path, but two bugs in db-reconcile.yml reproduce the exact silent-drift failure this PR is meant to close. Requesting changes on those two; rest can be follow-ups.

Blocking

.github/workflows/db-reconcile.yml:300 — all_present and the apply step's FILES list are still hardcoded to migrations 0010-0014, but the ledger-rebuild VALUES loop below now reads all of meta/_journal.json. Once 0015 lands, this workflow can insert a ledger row for it without ever checking or applying it — silently defeats the whole point of this PR on the very next migration.

.github/workflows/db-reconcile.yml:322 — the sha256sum call inside the while loop can fail (missing/renamed .sql file) without tripping set -euo pipefail — confirmed by reproducing it locally, a failing command substitution in a non-last pipe stage doesn't propagate. Result: an empty-hash row gets silently inserted instead of the workflow aborting. Needs an explicit [ -n "$hash" ] || exit 1 check per iteration.

Should fix

- import-seed-varieties.ts:38 / import-general-ima real ~940 rows doesn't actually guard against atruncated export; a CSV cut off at row 300 still clears the floor and mass-prunes real rows.
- Neither importer closes the pool createTargetDbruns (migrate-remote-db.ts does via pool.end();these don't) — process hangs past the pg idle timeout.

rcjasub (approved):
LGTM

PR #1006, opened 2026-07-04
fix(site): hide account nav link until onboarding is complete

### LINKED ISSUE #996: [Bug]: "Account" link broken during onboarding
### Bug Description

On the /apply flow (signup → bank information → terms), the top-right navigation shows an "Account" link. Clicking it goes to /account, but the user hasn't completed onboarding yet and has no account/farm set up, so the link doesn't work / leads nowhere meaningful. The link is rendered unconditionally in NavLinks (apps/site/src/components/common/authenticated-header/nav-links.tsx:19-22) via AuthenticatedHeader, which is shared across all routes under apps/site/src/app/(authenticated)/(chromeless)/(with-header)/ — including apply, welcome, and application-success — none of which are gated on onboarding completion.

### Steps to Reproduce

1. Sign up for a new account and begin the /apply flow.
2. Proceed through signup → bank information → terms (before onboarding is complete).
3. Look at the top-right of the header — click "Account."
4. See broken/non-functional destination since no account exists yet.

### Expected Behavior

The "Account" link should not appear (or should be disabled/hidden) until the user has an active account/farm. It should only show once onboarding is complete

### Screenshots/Videos

_No response_

### Browser

Chrome

### Device Type

Desktop

### Language/Locale

None

### Additional Context

_No response_

dgonz10152:
## Description

Fixes #996 

## Checklist

### DIFF
--- a/apps/site/src/components/common/authenticated-header/nav-links.test.tsx
+++ b/apps/site/src/components/common/authenticated-header/nav-links.test.tsx
@@ -1,14 +1,46 @@
 // Copyright © Todd Agriscience, Inc. All rights reserved.
 
 import { render, screen } from '@/test/test-utils';
-import { describe, expect, test } from 'vitest';
+import { beforeEach, describe, expect, test, vi } from 'vitest';
 import { NavLinks } from './nav-links';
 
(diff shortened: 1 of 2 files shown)


rcjasub (approved):
LGTM

PR #1007, opened 2026-07-07
Fix missing OpenAI embeddings key handling

### LINKED ISSUE #1004: Inference search shows "Something went wrong loading results. Please try again."
## Symptom

Submitting a query in the inference search panel (command palette → search) renders the red error state instead of results:

> Something went wrong loading results. Please try again.

Rendered by `apps/site/src/app/(authenticated)/components/search-panel/search-panel-body.tsx:86-90` whenever `searchError` is set.

## How the error state is reached

`search-panel-context.tsx:94-137` sets `searchError` when the `runInferenceSearch` server action rejects, or when no response arrives within 30s (`SEARCH_TIMEOUT_MS`). The action (`apps/site/src/app/(authenticated)/actions/inference-search.ts`) → `searchKnowledge` (`apps/site/src/lib/ai/search.ts`) throws in these cases:

1. **`getEmbedding(query)` throws** (`search.ts:107`) — this happens *before* any DB source runs, so it fails every search. `packages/db/src/utils/get-embedding.ts` reads `OPENAI_EMBEDDINGS_KEY` and falls back to the literal `'NOTAKEY'` when unset, so a missing/invalid key in the deployed Vercel environment guarantees a 401 from OpenAI and this exact error on every query. **Most likely cause if the error is 100% reproducible in one environment.**
2. **All three DB sources fail** (`search.ts:175-177`, "All search sources failed") — e.g. the database is unreachable or the shared `knowledge_article` join table is missing. Note PR #980's isolation (commit `1d0e8a9`) means a *single* missing table (like `general_imp` before the reconcile runs) only degrades that source; it should not trigger this error on its own.
3. **`getAuthenticatedInfo()` rejects** (`inference-search.ts:19`) — expired session.
4. **30s client timeout** — cold start + slow embedding + slow queries.

## Triage steps

- Check Vercel logs for the failing environment: `searchKnowledge` logs `[Search] Search failed:` and per-source `[Search] <label> source failed:` (`search.ts:92, 221`) — the logged error pinpoints which case above fired.
- Verify `OPENAI_EMBEDDINGS_KEY` is set for the affected Vercel environment (production/preview).
- Verify DB connectivity env vars for the same deploy.

## Related

- PR #1003 fixes the underlying migration-drift automation and adds the reconcile/import path for `general_imp` (rollout steps in the PR description).
- A distinct UX gap: the error copy is identical for auth expiry, config errors, and timeouts — worth splitting once the root cause is confirmed.

yjm7vj:
## Description

Fixes #1004

Removes the `NOTAKEY` fallback for `OPENAI_EMBEDDINGS_KEY` in the OpenAI embeddings client setup.

Previously, when `OPENAI_EMBEDDINGS_KEY` was missing, the app initialized the OpenAI client with the placeholder value `NOTAKEY`, causing inference search to fail later with a generic OpenAI/auth error. This change fails early with a clear missing environment variable error, making the root cause easier to diagnose in logs.

## Checklist

  - Not applicable; this is an environment variable guard/config handling change.
  - Not applicable; the thrown error message documents the required env var.
  - Not applicable; no UI components were modified.

### DIFF
--- a/packages/db/src/utils/get-embedding.ts
+++ b/packages/db/src/utils/get-embedding.ts
@@ -4,8 +4,14 @@ import OpenAI from 'openai';
 
 const OPENAI_API_KEY = process.env.OPENAI_EMBEDDINGS_KEY;
 
+if (!OPENAI_API_KEY) {
+  throw new Error(
+    'Missing OPENAI_EMBEDDINGS_KEY. Set it in apps/site/.env.local for local development and in Vercel environment variables for deployed environments.'
+  );
+}
+
 const openai = new OpenAI({
-  apiKey: OPENAI_API_KEY ?? 'NOTAKEY',
+  apiKey: OPENAI_API_KEY,
 });
 
 export async function getEmbedding(text: string): Promise<number[]> {


rcjasub (commented):
Requesting changes — the fix moves the missing-key check to module load time instead of scoping it to where the key is actually used, which introduces a regression.

packages/db/src/utils/get-embedding.ts now throws as soon as the module is imported, not when getEmbedding() is called. queries.ts statically imports getEmbedding, and its getDefaultDropdownItems() (lines 38–77) never calls it — it only queries IMPs/seed products directly. So the nav-bar's default (empty-query) search dropdown now hard-crashes whenever OPENAI_EMBEDDINGS_KEY is missing, even though that path never touched embeddings before. It also bypasses the graceful-degradation try/catch already in queries.ts/search.ts, since those only wrap runtime calls, not module evaluation.

Suggestion: move the check inside getEmbedding() (or lazily construct the client on first call), consistent with how connection.ts handles DATABASE_URL — checked lazily, at use time, not at import.

Otherwise the fix is correct and the error message is clear for #1004 — just needs to be scoped narrower before merging. Let me know how that goes.

vincenttodd (approved):

Jose-Gael-Cruz-Lopez:
## This is approved, but merging it would break CI on every PR and every release

Flagging before anyone acts on the approval — `main` has already solved this, and this change is the specific shape that was tried and deliberately reverted.

### It is superseded

`packages/db/src/utils/get-embedding.ts` on `main` (commit `7834283`, #1065) is now a lazy accessor performing the **identical** check with the **identical** error string this PR proposes:

```ts
function embeddingsClient(): OpenAI {
  const apiKey = process.env.OPENAI_EMBEDDINGS_KEY;
  if (!apiKey) {
    throw new Error('Missing OPENAI_EMBEDDINGS_KEY. Set it in apps/site/.env.local …');
  }
  client ??= new OpenAI({ apiKey });
  return client;
}
```

Its JSDoc records why it is lazy rather than module-scope:

> A module-scope throw took those scripts down wherever no embeddings key is configured — including the Database Compatibility Check […] The check itself is unchanged, only its timing.

### Why the module-scope throw breaks CI — full static trace

This PR adds the throw at module scope, which fires at **import**, not at use:

1. `.github/workflows/pr-database-compatibility-check.yml:94` runs `bunx tsx scripts/migrate-remote-db.ts --env staging`
2. `.github/workflows/release-database-compatibility-check.yml:92` runs the same script with `--env prod`
3. `packages/db/scripts/migrate-remote-db.ts:21` → `import { getArg } from './lib/importer-lib'` (module-level, and it only wants a CLI-flag helper)
4. `packages/db/scripts/lib/importer-lib.ts:12` → `import { getEmbedding } from '../../src/utils/get-embedding'` (module-level)
5. `OPENAI_EMBEDDINGS_KEY` appears **exactly once** across all of `.github/workflows` — at `sync-imps-to-staging.yml:19`. **Neither** database-compatibility job carries it.

So loading the migrate script executes this module unconditionally, with no key present, on every PR and every release.

Note the import at `importer-lib.ts:12` cannot simply be removed either — `importer-lib.ts:149` genuinely calls `getEmbedding(text)` for real embedding callers. Deferring the check into the function is the only shape that keeps both the migrate path and the embedding path working, which is exactly what `main` now does.

### Blob-hash evidence that this exact state already lived on main

- This PR's diff header: `index 282085ef..5595b392`
- #1063's diff header for the same file: `index 5595b3920..282085efc` — the exact reverse

So `5595b392` (the module-scope-throw blob this PR creates) was on `main` at one point, and `main` has since moved past it to the lazy client. The timeline is `NOTAKEY` → module-scope throw → lazy `embeddingsClient()`.

### Also

It would contradict `get-embedding.test.ts:21-28` ("imports without a key configured"), turning the `packages/db` suite red.

### Recommendation

Close as superseded by #1065. The same hunk appears in #1012 and #1021 and should be dropped from both; #1063 reverts the file in the other direction, to `apiKey ?? 'NOTAKEY'`, which is also worse than `main` (it converts a named configuration error into a silent 401 at call time).

<sub>Automated review. Every path above was read at `origin/main` = `a688cfe`; no code was changed by this comment.</sub>

Jose-Gael-Cruz-Lopez:
Closing: this is already solved on `main`, and merging it now would reintroduce the bug it was meant to fix.

#1065 replaced the module-scope check with a lazy `embeddingsClient()`. The reason is recorded in its JSDoc: `get-embedding.ts` sits on the import path of scripts that never embed anything — `migrate-remote-db.ts` reaches it through `importer-lib` purely for a CLI-flag helper. A module-scope `throw` fires at **import**, which took down both `pr-database-compatibility-check.yml` and `release-database-compatibility-check.yml`; neither has `OPENAI_EMBEDDINGS_KEY` (it appears once across all workflows, in `sync-imps-to-staging.yml`).

This branch adds exactly that module-scope `throw` back. It also diffs against a version of the file that no longer exists — `main` has no `?? 'NOTAKEY'` any more — so it conflicts.

The check itself was kept, only its timing changed: anything that actually requests an embedding without a key still fails, and fails by name. Thanks for raising it, the diagnosis was right even though the fix moved.

PR #1009, opened 2026-07-09
fix(site): dedupe getauthenticatedinfo db queries per request

### LINKED ISSUE #1008: [Bug]: NavLinks duplicates getAuthenticatedInfo() DB query on every authenticated page load
### Bug Description

On every page under `(authenticated)` that renders `AuthenticatedHeader`, `getAuthenticatedInfo()` runs twice per request: once in `ViewerAgreementGate` (apps/site/src/app/(authenticated)/layout.tsx:23), and again in `NavLinks` (apps/site/src/components/common/authenticated-header/nav-links.tsx:19), added by the #996 fix that hides the "Account" link until onboarding is complete.

`getAuthenticatedInfo` runs a `user ⋈ farm` join query and is a plain `async function` — it is not wrapped in React's `cache()`, so there is no request-level memoization; each call is a fresh DB round trip. For users who haven't finished onboarding, `NavLinks` also calls `hasCompletedPlatformOnboarding` → `hasAcceptedAccountAgreement`, a third query, which duplicates a check `ViewerAgreementGate` already performs for Viewer-role users.

Net effect: every authenticated page render now does 1-2 extra DB queries it didn't need to, site-wide.

### Steps to Reproduce

1. Add query logging (or a breakpoint) inside `getAuthenticatedInfo` (apps/site/src/lib/utils/get-authenticated-info.ts:19).
2. As an authenticated user, load any page under `(authenticated)` that renders `AuthenticatedHeader` (e.g. `/apply`, `/account`, `/v`).
3. Observe `getAuthenticatedInfo`'s user+farm join query execute twice for the single request: once from `ViewerAgreementGate` in the root layout, once from `NavLinks`.

### Expected Behavior

getAuthenticatedInfo` should resolve once per request and be reused by every caller in that request tree (e.g. 
wrapped in React's `cache()`, the standard Next.js App Router pattern for this), instead of issuing a duplicate DB round trip per caller.

### Screenshots/Videos

_No response_

### Browser

Chrome

### Device Type

Desktop

### Language/Locale

None

### Additional Context

Introduced by commit 7e0e0dd ("fix(site): hide account nav link until onboarding is complete", branch fix/996-hide-account-link-onboarding). Flagged as non-blocking during review of that PR — not a correctness issue, just added DB load. Suggested fix: wrap `getAuthenticatedInfo` in React's `cache()` from `react` so repeated calls within the same server request dedupe automatically.

dgonz10152:
## Description

Fixes #1008 

## Checklist

### DIFF
--- a/apps/site/src/lib/utils/account-agreement.ts
+++ b/apps/site/src/lib/utils/account-agreement.ts
@@ -3,19 +3,22 @@
 import { accountAgreementAcceptance } from '@nightcrawler/db/schema';
 import { db } from '@nightcrawler/db/schema/connection';
 import { eq } from 'drizzle-orm';
+import { cache } from 'react';
 
 /** Returns whether the given user has accepted the account agreement.
  *
  * This checks persisted acceptance state from the database.
+ * Wrapped in React's `cache()` so repeated calls within the same server
+ * request resolve once instead of issuing duplicate DB queries.
  */
-export async function hasAcceptedAccountAgreement(
-  userId: number
-): Promise<boolean> {
-  const [hasAccepted] = await db
-    .select({ id: accountAgreementAcceptance.id })
-    .from(accountAgreementAcceptance)
-    .where(eq(accountAgreementAcceptance.userId, userId))
-    .limit(1);
+export const hasAcceptedAccountAgreement = cache(
+  async (userId: number): Promise<boolean> => {
+    const [hasAccepted] = await db
+      .select({ id: accountAgreementAcceptance.id })
+      .from(accountAgreementAcceptance)
+      .where(eq(accountAgreementAcceptance.userId, userId))
+      .limit(1);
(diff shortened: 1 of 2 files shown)


rcjasub (commented):
Requesting changes — this only fixes part of what #1008 reported.

getAuthenticatedInfo() is now correctly wrapped in cache(), so that duplicate query is gone.

But the bug also flagged hasAcceptedAccountAgreement() running twice for Viewer-role users with an unapproved farm — once in ViewerAgreementGate (layout.tsx:29), once via NavLinks → hasCompletedPlatformOnboarding (platform-onboarding.ts:23). That function (account-agreement.ts:11) wasn't touched here and still isn't memoized, so that duplicate remains.

rcjasub (approved):
LGTM

PR #1010, opened 2026-07-09
fix(site): cut sanity content staleness from an hour to five minutes

### LINKED ISSUE #900: [Bug]: Sanity/Nextjs cache delay
### Bug Description

Updates from Sanity are possibly being cached on the site, and followup edits in Sanity do not immediately update to the live site. Theres a very high chance we are caching `true` when it should be `false`

### Steps to Reproduce

1. Update something on Sanity
2. Publish
3. go to changed article

### Expected Behavior

It should update within a few minutes.

### Screenshots/Videos

_No response_

### Browser

Chrome

### Device Type

Desktop

### Language/Locale

None

### Additional Context

_No response_

henryng15:
## Description

Fixes #900

Published Sanity edits could take over an hour to show up on the live site. Two cache layers were stacked on top of each other:

1. **`useCdn: true`** in `lib/sanity/client.ts`. Reads went through the Sanity CDN, which serves its own cached responses and which Next.js has no way to invalidate.
2. **A one hour ISR window** on every fetch — `LISTING_REVALIDATE` in `articles.ts` and `FORM_REVALIDATE` in `forms.ts`, both `60 * 60`.

There was also a third, quieter problem. `getArticleBySlug` forwarded a bare `options` argument with no default, so each caller supplied its own window. Six page and layout files had `next: { revalidate: 60 * 60 }` copy-pasted inline. Changing the constant in `articles.ts` would not have touched a single article detail page.

### What changed

- `useCdn: false` — read from the Sanity API so Next.js alone governs staleness.
- A single `SANITY_REVALIDATE` (five minutes) and `defaultSanityFetchOptions`, both exported from `lib/sanity/client.ts`.
- `getArticleBySlug` and `getFormBySlug` now fall back to that default instead of passing `options` through untouched.
- The six inline `60 * 60` windows are gone; those pages inherit the default.

Callers that deliberately opt into their own caching are untouched: `sitemap.ts` keeps its one day window, and the `/forms/[slug]` server action keeps `cache: 'no-store'`.

### Tradeoff

This swaps CDN reads for API reads, which count against the Sanity project quota. Current marketing traffic is well inside that budget. If it ever isn't, the durable fix is a Sanity webhook calling `revalidateTag` on publish — that would let us raise the ISR window back up while making publishes appear immediately. It needs admin access on the Sanity project and a new secret, so it belongs in its own issue.

### Testing notes

Three new co-located test files under `lib/sanity/`. Each was verified to fail against the pre-fix code:

- `client.test.ts` — asserts `useCdn` is off and pins `SANITY_REVALIDATE` at five minutes or less, so a future edit cannot quietly restore the hour.
- `articles.test.ts` / `forms.test.ts` — assert the default fetch options reach `client.fetch` when a caller passes none, and that an explicit override still wins.

`bun validate` passes end to end: `format:check`, `type-check`, `lint`, `test`, `build`.

Not verified locally: the actual propagation delay against the live Sanity dataset. That needs a publish in Sanity Studio and a look at the deployed preview.

## Checklist


rcjasub (approved):
based on everything I verified (diff matches the description exactly, tests pass, type-check/build clean, no orphaned 1-hour caches left anywhere, branch is current with main) LGTM

PR #1011, opened 2026-07-10
chore: repo cleanup — dead code, single-source docs, kebab-case naming

Jose-Gael-Cruz-Lopez:
## Description

No linked issue — this is a housekeeping branch.

A full cleanup pass over the monorepo: remove dead code and unused dependencies, install and configure `knip` so this stays measurable, consolidate four drifting convention documents into `AGENTS.md` as the single source of truth, fix the logging level regression it exposed, and enforce the kebab-case filename rule that `AGENTS.md` has always specified.

**111 files changed, +925 / −2452** across 12 commits. 29 files deleted, 30 renamed, 18 dependencies removed, 1 added (`knip`).

**No behavior change except one, called out explicitly below.**

---

### Dead code and dependencies

| Commit | What |
|---|---|
| `1f03901` | Install `knip` + `knip.json` + root `knip` script. Delete 20 unreferenced files. Remove 17 unused deps. |
| `cc49253` | Delete two superseded header components (7 files, −414 lines). |
| `8b5a376` | Rename `new-header/` → `header/` now that the old name is free. |
| `18e6b2e` | Cascade sweep: delete `ui/drawer.tsx` + `vaul`, delete `lib/types/components.ts`. |

Two marketing/campaign headers had been replaced but never pruned. `(marketing)/layout.tsx` renders `NewHeader`; the old `header/` folder had **zero production references**. Same story for `components/go/header/`, superseded by `(go)/creators/components/header/`.

Neither was visible to `knip`: test and story files are entry points, so **a dead component with a passing test still looks reachable**. Both folders were imported solely by their own `.test.tsx` and `.stories.tsx`. `header.test.tsx` was contributing 8 green test cases for a component no route rendered.

Deleting the old header cascaded: it was the sole consumer of `ui/drawer.tsx` (itself `vaul`'s only consumer) and of `lib/types/components.ts` (a single 4-line `MenuItem` interface the new header replaced with its own). The shadcn drawer primitive can be re-added at any time.

**Dependencies removed (18):**
- `apps/site` — `@dnd-kit/core`, `@dnd-kit/sortable`, `@dnd-kit/utilities`, `@sanity/vision`, `ua-parser-js`, `@types/ua-parser-js`, `embla-carousel-fade`, `intersection-observer-polyfill`, `preact`, `react-docgen`, `undici`, `vaul`, `@chromatic-com/storybook` (dev)
- `apps/internal` — `@hookform/resolvers`, `@radix-ui/react-dropdown-menu`, `@radix-ui/react-tooltip`, `zod`
- `packages/db` — `drizzle-seed`

Kept despite `knip` flagging them, because they are real: `newrelic` (loaded via `require()` in `newrelic-agent.cjs`) and `tw-animate-css` (an `@import` in `globals.css` — `knip` scans neither). Both are recorded in `knip.json` as `ignoreDependencies`.

### Documentation — one source of truth

| Commit | What |
|---|---|
| `2f9c00d` | `AGENTS.md` becomes the single source of truth (40 → 224 lines). `CLAUDE.md` reduced to site-specific (287 → 183) and links back. All 5 READMEs reconciled against `package.json`. |
| `dd49773` | `.agent/instructions.md` + 7 workflows reduced to pointers; genuine runbook content preserved. |
| `66f14de` | `.cursorrules` reduced 278 → 29 lines. `CONTRIBUTING.md` jest → vitest. |
| `a44459a` | `CODEOWNERS`: 6 stale paths repointed, copyright header corrected. |

There were **four** convention documents, and where they overlapped they disagreed. Contradictions resolved:

- **Naming.** `AGENTS.md` said all files kebab-case; `CLAUDE.md`/`.cursorrules` said "kebab-case for directories, PascalCase for components." These only looked irreconcilable — `CLAUDE.md`'s own example shows `hero.tsx` exporting `Hero`, so it meant the *identifier*. Now stated as: **filenames kebab-case, identifiers PascalCase.**
- **State.** `AGENTS.md` said prefer `useForm()`; `CLAUDE.md` said use `useState`/`useReducer`. Split by scope: `useForm` for form state, `useState`/`useReducer` for non-form local state, Context for cross-cutting.
- **Logging.** `CLAUDE.md` forbade `console.log` in one section and permitted it in another, four lines apart.

Stale facts corrected, each verified against the repo: `cd frontend` named a directory that does not exist (it appeared in `AGENTS.md`, `.agent/instructions.md`, `.agent/workflows/commit-and-push.md`, and `.cursorrules`); `bun ci` and `bun lint:fix` exist only in `apps/site`, not at the root; **pre-commit does not type-check** (that runs on pre-push); Next.js 15 → 16.3, Storybook 9.1 → 10.3; `CLAUDE.md`'s directory tree referenced five paths that no longer exist; four bullets were duplicated verbatim.

`.cursorrules` also told you to write `"Copyright Todd LLC, All rights reserved"` as the file header. The enforced string is `// Copyright © Todd Agriscience, Inc. All rights reserved.`, so anyone following it produced a lint failure on every new file.

`CODEOWNERS` had six rules pointing at paths that don't exist, so those files silently fell through to the `*` catch-all: `prettier.config.*` → `.prettierrc`/`.prettierignore`; `jest.config.js`/`jest.setup.*` → the real vitest configs; `apps/site/src/middleware.ts` → `src/proxy*` (Next 16 renamed it); `package-lock.json` → `bun.lock`; `envexample.md` → `**/env.example`. `/AGENTS.md` now has an explicit owner. Jest is not a dependency of any workspace.

### Logging

| Commit | What |
|---|---|
| `c042693` | Route `console.*` through the workspace logger. |
| `2f6a3f5` | **Behavior change** — see below. |

Excluding `/scripts` (intentional CLI output), test files, and the three `logger.ts` implementations, exactly one `console.*` call remained in app/library source: `apps/site/src/i18n/request.ts`.

### `imp` form type

| Commit | What |
|---|---|
| `8c1dfc5` | Export `ImpNotesFormValues`, import it explicitly, drop the `knip` ignore. |

`imp/[slug]/components/types.ts` had **no import or export statement**, making it a global script whose `interface ImpNotesFormValues` leaked into the ambient type namespace for the entire project. `imp-notes-form.tsx` consumed it without importing anything. That fooled every tool that reasons about the module graph — `knip` reported the file as unused, and a grep for exported symbols found none. Only `tsc` noticed when it was deleted. `apps/site` now has **zero** global-script files.

### Naming — kebab-case, two waves

| Commit | What |
|---|---|
| `18e6b2e` | Wave 1 — 12 renames across `apps/site` source, `.storybook`, and `apps/sanity`. |
| `bd71a4e` | Wave 2 — the two i18n message directories. |

**No exported identifier was renamed in either wave.** `ThemeProvider`, `useTheme`, `ThemeContextType`, `ThemeProviderProps`, `submitContactToSheets`, `useCookiePreferences`, `useCurrentUrl`, `useWindowWidth`, `withFonts`, `withStorybookProvider`, `storybookControls` and `schemaTypes` all keep their names. Only filenames changed. (`8b5a376` did rename `NewHeader` → `Header`, deliberately, as part of finishing the header swap.)

`apps/site/tsconfig.json:31` follows the `storybook-controls` rename. That `include` entry is load-bearing: TypeScript's `**/*.ts` glob does not traverse dot-directories, so without it the file drops out of the program. Verified with `tsc --listFiles` before and after. **It is the only config line changed on this branch.**

**Wave 2 deserves a note.** The directory name is *not* the namespace. `request.ts` builds a dynamic import path from the `messageFiles` array and merges each JSON with `Object.assign`, so the namespace is the JSON's own top-level key. Only path-shaped references changed. The six `getTranslations`/`useTranslations` namespace strings, the `enMessages.articleIndex` property access, and all four JSON top-level keys stay camelCase — they are identifiers, not filenames.

A passing build proves nothing there, so I ran a negative control: pointing `messageFiles` back at the old name made `cookie-preferences-modal.test.tsx` fail with 20 literal `[cookiePreferences.title]`-style fallbacks. The loader's `console.warn` never fires under vitest (Vite rewrites the dynamic `require`), so **the fallback literal is the only signal**. `articleIndex` has no test that consumes the mocked translator, so tests alone could not prove it — the cold production build does: prerendered output contains `"Newsroom"`, `"Cookie Settings"` and `"Your Privacy Choices"`, with zero `[namespace.key]` fallbacks.

---

## ⚠️ The one behavior change

`apps/site/src/i18n/request.ts` previously logged a failed message-file load at **error** level in production. It now logs at **warn**, which is development-gated, so **nothing is emitted in production**.

This is intentional. Message JSON is bundled at build time, so a missing translation file is a build-time concern, not a live-request event — and error-level would page anyone alerting on error logs. All 18 namespaces have both `en.json` and `es.json`. The `"Warning:"` prefix is gone; it only existed to compensate for misusing the error channel. `logger.warn` is already dev-gated, so the surrounding `NODE_ENV` branch collapsed into a single call. **No new log level was added** — the logger keeps its errors-only-in-production philosophy.

---

## Verification

Run cold, as one unit, against the tip — `rm -rf apps/site/.next` then `TURBO_FORCE=true bun run validate`. **12 tasks force-executed, `Cached: 0 cached`, `BUILD_ID` regenerated.** A cached green would be meaningless.

| Stage | Result |
|---|---|
| `format:check` | clean |
| `type-check` | **4/4** workspaces |
| `lint` | **0 errors**, 2 warnings (pre-existing, see below) |
| `test` | **517 passed** (506 site + 11 db), 0 failures |
| `build` | **3/3** (site, internal, sanity) |

**`bun run build-storybook` — exit 0**, run separately because `validate` provably cannot see the `.storybook` tree (not in the TS program, `globalIgnores`d by ESLint, no turbo task). Zero module-resolution errors, and a string unique to the renamed `with-storybook-provider.tsx` appears in the built bundle — proving the decorators resolved rather than being silently skipped.

**`bun run knip`** — 46 unused files / 10 deps / 4 devDeps / 1 unlisted / 106 exports / 42 types / 11 duplicates. Every one of the 46 is an intentionally deferred tree (farm+GBI 23, tabs 6, widgets 15, `ui/` 2). Nothing unaccounted for.

**Nothing was loosened.** `eslint.config.mjs`, `.prettierrc`, `.prettierignore`, `turbo.json`, `commitlint.config.cjs`, three of four `tsconfig.json`, and all three husky hooks are **byte-identical** to the branch point. **Zero `eslint-disable` / `@ts-ignore` / `@ts-expect-error` added to any source file.**

The 2 lint warnings are `@next/next/no-img-element` at `(marketing)/index/[slug]/page.test.tsx:17,24`, in a file byte-identical to the branch point, where mocking `next/image` with a raw `` is exactly right. I did not add an override.

I also ran the app: `/`, `/es`, `/research/index`, `/news` and `/es/research/index` all return 200, render the renamed header and theme context, resolve both locales, and emit zero `[namespace.key]` fallbacks.

---

## Needs a human before merge

**Smoke-test the authenticated header.** The five-file `authenticated-header/components/` search cluster deleted in `1f03901` sits behind auth. `not-found.test.tsx` mocks `AuthenticatedHeader` rather than rendering it, so **no test exercises the real component**. Static reachability proved zero importers and the whole gate is green, but this is the one deletion no automated check covers.

---

## Deliberately not in this branch

- **`apps/sanity` has no `lint` script**, so `turbo lint` skips it — which is why nobody noticed that `eslint-plugin-react@7.37.5` crashes under `eslint@10.5.0` on any Sanity `.ts` file. `18e6b2e` needed `--no-verify` because of it. Pre-existing; the fix is a `lint` script using Sanity's own config.
- **No `.gitattributes`.** `CODEOWNERS` had mixed CRLF/LF line endings; `a44459a` normalized that one file to LF, but the root cause remains.
- **Storybook is not in CI** — `turbo.json` has no `build-storybook` task.
- **Two dead-but-tested trees remain**: `common/news/{featured-news-carousel,latest-news-table}.tsx` and `investors/governance/components/team-member-card/`. Same failure mode as the headers; `knip` cannot see them.
- **JSDoc coverage is 76%** (721/944 exported declarations). Reported, not acted on — a quarter of the gap is `metadata` boilerplate in `layout.tsx` files and drizzle type re-exports.
- **`bun run generate-icons` is broken** — `apps/site/package.json` declares `tsx scripts/generate-icons.ts`; that file does not exist.
- **8 stale `"Copyright Todd LLC"` headers** remain in `.devcontainer/` and `.github/`. Only `CODEOWNERS` was in scope.
- **`.github/CONTRIBUTING.md` still duplicates conventions** (JSDoc, logger, conventional commits, license header) without pointing at `AGENTS.md`. It does not contradict it. Only the jest→vitest line was in scope.

---

## Checklist

- [x] You've included unit or integration tests for your change, where applicable. — *No new behavior to test. The suite went **527 → 517**: exactly the 10 tests that exercised the two dead headers were deleted along with them. Correctness is proven by a cold `validate`, a separate Storybook build, a negative control on the i18n rename, and a live dev-server smoke test.*
- [x] You've included inline docs for your change, where applicable. — *`AGENTS.md` rewritten as the single source of truth; `CLAUDE.md`, `.cursorrules`, `.agent/*` and all 5 READMEs reconciled against the actual `package.json` scripts.*
- [x] Any components that you've modified are accessible. — *No component markup changed. `8b5a376` is a pure rename; the rendered header is byte-identical.*
- [x] You've used [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/) where appropriate. — *All 12 commits pass `commitlint`.*


vincenttodd (commented):
looks good to me

apps/site/src/app/(unauthenticated)/[locale]/(marketing)/components/header/header.tsx
```diff
@@ -2,148 +2,43 @@
...
-    { href: '/research', label: t('navigation.whatWeDo') },
-    { href: '/news', label: t('navigation.news') },
-    { href: '/careers', label: t('navigation.careers') },
-  ];
+export default function Header({
+  menu,
+  auth,
+  className,
+}: HeaderProps): React.ReactNode {
+  const { menuItems, authLinks } = useMenuItems({ menu, auth });
```
Comment (kurealnum) [inline apps/site/src/app/(unauthenticated)/[locale]/(marketing)/components/header/header.tsx:29]: not rlly sure what this is doing, seems like an antipattern from a brief glance

apps/site/src/components/common/authenticated-header/components/search-dropdown.tsx
kurealnum, line 1: i might just be blind asf but where did all of these go/get moved to 😭 or is this unused code now

AGENTS.md
kurealnum, line 1: might wanna try and minimize this, this is like 2.5k tokens that's loaded on every startup of Claude/your agent (not that bad but still, could probably minimize). separating these into skills/subagent personas might be a good idea. ex. have `./claude/agents/frontend-agent.md` with all info about making a frontend change in there and put "If you make a frontend change, use this persona", etc.

kurealnum (commented):
3 minor things, nothing big

vincenttodd:
wait lowkey good idea if we were gonna make different personas, colossus has some frontend context files, that, if someone fixed my countless typos, could use that

Jose-Gael-Cruz-Lopez:
## Rebase assessment — the analysis still holds, but `share-article` needs a decision

I simulated the merge with `git merge-tree HEAD origin/main` (no working tree touched). Results below.

### Correction to an earlier note on this PR

I previously suggested this PR deletes `share-article.tsx` while leaving `share-article.test.tsx` orphaned. **That was wrong**, and the real situation is more interesting:

- At the merge base (`9b1daf2`) only `share-article.tsx` existed — the test did not
- `main` has since **actively worked on it**: `23b891c` (#1020, "repair LinkedIn share URL + add UTM to share buttons") modified the component and added a 92-line test

So it is a **modify/delete conflict**, not a silent orphan — git will stop and force a human decision rather than producing a broken tree:

```
CONFLICT (modify/delete): apps/site/src/components/common/share-article/share-article.tsx
deleted in HEAD and modified in origin/main.
```

The dead-code call is still defensible — `share-article` has **no external importers on `main`** (verified). But deleting it now discards #1020's bug fix and its test. Worth a conscious call: either drop it from this PR and file a separate "remove unused share-article" ticket referencing #1020, or delete it and note that #1020's work is being dropped deliberately.

### Full conflict list from the merge simulation

**Content conflicts (6)**
- `apps/site/.storybook/preview.ts`
- `apps/site/src/app/(unauthenticated)/[locale]/(contact)/contact/page.tsx`
- `apps/site/src/app/(unauthenticated)/[locale]/(marketing)/about/components/competencies-section/competencies-section.tsx`
- `apps/site/tsconfig.json`
- `bun.lock`
- `package.json`, `packages/db/package.json`

**Modify/delete conflicts (4)** — these need explicit decisions, not just merge-tool resolution
- `contact/actions.ts` and `contact/actions.test.ts` — deleted on `main` (Sanity migration), modified here
- `components/header/header.stories.tsx` — deleted here, modified on `main`
- `components/common/share-article/share-article.tsx` — deleted here, modified on `main` (see above)

Everything else auto-merges. **The expensive part of this PR — the 29 deletions, 30 renames and ~20 importer rewrites — is still valid work.** The conflicts are concentrated in config/lockfile plus these four decisions.

### Two things worth pulling out of the "chore" framing

1. **19 runtime dependencies are removed** (`@dnd-kit/*`, `preact`, `ua-parser-js`, `undici`, `vaul`, `@sanity/vision`, `drizzle-seed`, …). All verified unused, but that is the highest-blast-radius part of the PR and deserves a focused reviewer rather than being read as naming cleanup.
2. **`apps/site/src/i18n/request.ts` swaps `console.warn` → `logger.warn`.** `logger.warn` is gated on `shouldLog()` (`lib/logger.ts:14-36,60-64`), which is false in production — so a missing message file now fails **silently** in prod where it previously logged. Defensible and it does follow the AGENTS.md logging rule, but it is a production behaviour change inside a 111-file chore.

### Also

- `apps/site/tsconfig.json` — this hunk is now **obsolete**; `main` already generalised the Storybook include to `.storybook/**/*.ts{,x}` globs. Drop it on rebase.
- Zero case-only renames, so no macOS case-insensitivity hazard — the rename half can be one mechanical commit.
- Header consolidation deletes `header.stories.tsx` and `header.test.tsx` without replacements for the promoted component, dropping the banner-landmark and menu-toggle a11y assertions from `header.test.tsx:14-23`. `.github/CONTRIBUTING.md` requires stories for all UI components.

### Suggested split

Risk rises monotonically and each piece is independently revertible:

1. **Docs consolidation** (~19 files) — `AGENTS.md`'s pre-image is byte-identical to `main`, merges today with no rebase
2. **Dead code + `knip.json` + the 19 dependency removals** — carries the real risk, wants a focused review
3. **Header consolidation** — a semantic change, not a chore; needs its own title and replacement story/test
4. **Kebab-case renames** — land last, in an announced window, since they invalidate every other open PR's paths

<sub>Automated review against `origin/main` = `a688cfe`. Merge simulated with `git merge-tree`; no branch or working tree was modified.</sub>

akeight:
This might be done with the merging of the 3 PRs earlier. Needs verification before closing.

Jose-Gael-Cruz-Lopez:
Closing: superseded by #1089, #1090 and #1091, which all merged. Splitting this into docs / dead-code+knip / kebab-case made each piece reviewable; at 111 files with 11 conflicts this branch was not landable as one unit.

PR #1017, opened 2026-07-12
fix(site): improve error handling for /contact

### LINKED ISSUE #740: Improve error handling for /contact
## Problem Statement
The `/contact` flow needs stronger error handling so failures are surfaced clearly and users can recover without confusion. The current behavior of the error message on this page is to show the error from the server, which is confusing and unhelpful.

## Context
`/contact` is a primary entry point for inbound communication. If submissions fail silently or with unclear messaging, users may abandon the flow or submit duplicate requests.

## Constraints
This issue should not:
- Introduce breaking API changes for existing contact submission integrations
- Leak internal error details to end users
- Regress successful submission UX

## Technical Details
Implement and standardize error handling across the `/contact` path:
- Handle client-side validation errors with generic error message

## User Story
As a visitor submitting the contact form, I want clear feedback when something goes wrong so that I can correct issues or retry confidently.

## Use Case
- User fills out and submits `/contact`
- Submission fails due to validation/network/server condition
- UI shows a generic error message

BaoT1301:
## Description

Fixes #740

The public inquiry action on `/contact` forwarded the caught error's raw message directly to the client:

```ts
throwActionError(err instanceof Error ? err.message : 'An unknown error occurred.');
```

That can leak internal details to end users (upstream HTTP status codes, a missing `CONTACT_GOOGLE_SCRIPT_URL` config error, or other exception text that should never leave the server).

**What changed**
- The full error is still logged server-side with `logger.error` for diagnosis.
- The caller now always receives a fixed, generic message ("We could not submit your inquiry right now. Please try again later."), regardless of the underlying failure.

**Tests**
- Updated the failure-path test to assert the generic message is thrown, that the raw internal message (`Sheets is down`) is not surfaced to the caller, and that the full internal error is still logged server-side.

No UI or behavior change beyond the error text the user sees on a failed submission.

## Checklist


Jose-Gael-Cruz-Lopez:
## Both files this PR edits no longer exist, and the fix it proposes is already on main

### Why all four checks fail

This PR modifies:

- `apps/site/src/app/(unauthenticated)/[locale]/(contact)/contact/actions.ts`
- `apps/site/src/app/(unauthenticated)/[locale]/(contact)/contact/actions.test.ts`

Neither exists on `main`. Verified directly:

```
$ git cat-file -e origin/main:'apps/site/src/app/(unauthenticated)/[locale]/(contact)/contact/actions.ts'
fatal: path '…/contact/actions.ts' does not exist in 'origin/main'
```

The `(contact)/contact/` directory on `main` now contains only `layout.tsx`, `layout.test.tsx`, `page.tsx`, `page.test.tsx`. A grep for `submitPublicInquiry` across `apps/site` returns **zero matches** — the symbol is gone.

That is a **modify/delete conflict on 2 of 2 files**, so no mergeable ref can be produced, and `code-quality.yml`, `testing.yml` and `build.yml` all fail together against it. That matches the observed "all four red" signature better than any code-level defect.

Contact was migrated to the Sanity form builder — `contact/page.tsx:14-23` now fetches `getFormBySlug('contact')` and renders `<DynamicForm />`.

### The behaviour you wanted is already implemented

`apps/site/src/app/(unauthenticated)/[locale]/(marketing)/forms/[slug]/actions.ts:102-103`:

```ts
} catch (error) {
  logger.error('Form submission failed', error);
  throwActionError('Unable to submit your request. Please try again.');
}
```

Log the real error, return a generic one — the same pattern this PR proposed, now applied to **all** Sanity-driven forms including contact, rather than to one action.

### The original concern was valid

Worth recording, because the instinct was right: `googleSheets.ts:85` throws `'Server configuration error: Missing CONTACT_GOOGLE_SCRIPT_URL'`, and the old `err instanceof Error ? err.message : …` would have forwarded that string to the browser. Genuine (low-severity) information disclosure, correctly identified — just no longer reachable.

### Recommendation

Close as superseded by the Sanity form-builder migration.

### Unrelated follow-up this surfaced

`throwActionError` double-logs: `apps/site/src/lib/utils/actions.ts:64` already calls `logger.error`, and ~15 call sites log immediately before calling it. That is why the original test asserted `toHaveBeenCalledTimes(2)`. `main`'s `forms/[slug]/actions.ts:102-103` has the same redundancy. Probably worth its own ticket to make `throwActionError` the single logging point.

PR #1018, opened 2026-07-12
feat(site): add reusable social links component

### LINKED ISSUE #786: Create socials link component
### Problem Statement

This issue will build a resuable commom component for the social links that are used in the landing footer and go footer (anywhere it's currently used in the site).

### Context

_No response_

### Constraints

_No response_

### Technical Details

_No response_

### User Story

_No response_

### Use Case

_No response_

BaoT1301:
## Description

Fixes #786

The marketing footer and the go/creators footer each hardcoded their own copy of Todd's social profile URLs, labels, and icons. That duplication meant a profile change (or a new platform) had to be made in two places and could drift, which it already had: the two footers listed different platform sets, and the go footer's Discord URL carried a trailing space.

**What changed**
- New `SocialLinks` component under `components/common/social-links`, with a single `SOCIAL_LINKS` source of truth for every profile (href, aria-label, icon).
- It renders accessible external icon links that open in a new tab with `rel="noopener noreferrer"`, and accepts optional `platforms` (which set and in what order) and `iconSize` props so each surface controls its own presentation without duplicating the data.
- Both footers now render `<SocialLinks />` instead of their own arrays.

**Kept identical on purpose**
- Each footer keeps its existing platform set and order (marketing: Instagram, LinkedIn, X, YouTube; go: X, Instagram, LinkedIn, YouTube, Discord), so the visible icons are unchanged. I did not silently add Discord to the marketing footer. If we'd prefer the two footers to show the same set, that's now a one-line change, happy to do it if you want.

**Only behavior change**
- The marketing footer's social links now open in a new tab (`target="_blank"` + `rel="noopener noreferrer"`), matching what the go footer already did. Previously they opened in the same tab via the locale-aware `Link`, which is meant for internal routes, not external URLs.

**Tests / stories**
- Unit tests: default render, platform subset and order, and new-tab safety.
- Storybook story with an all-platforms variant plus per-footer variants.

## Checklist


Jose-Gael-Cruz-Lopez:
## Cause of the failing checks, plus a footer hunk that no longer applies

### Build Application + Code Quality — attributed

`social-links.stories.tsx:3`:

```ts
import type { Meta, StoryObj } from '@storybook/nextjs-vite';
```

`main` migrated to `@storybook/react-vite` (`bca54ad`). On `main` today, `apps/site/package.json` devDependencies contain `@storybook/react-vite: ^10.3.5` and **no `nextjs-vite`**, which has **zero occurrences in `bun.lock`**. `.storybook/main.ts:3` and `preview.ts:3` both import `react-vite`, as does every existing story.

Because `apps/site/tsconfig.json:25-33` includes `**/*.tsx`, both `tsc --noEmit` (`code-quality.yml:33`) and `next build` / `build-storybook` (`build.yml:32,35`) resolve the story import and fail with TS2307. `eslint.config.mjs:49` globally ignores `*.stories.*`, so `lint` stays green — which is what makes the failure look strange.

One-word fix: `@storybook/react-vite`.

### Unit Tests + Security Audit — not attributable to this diff

Ruled out with evidence, in case someone chases them:

- Vitest never collects `.stories.tsx`, and `apps/site/vitest.config.ts:15-18` sets no `coverage.include`, so the bad import is invisible to the test run. Both footer tests assert socials by accessible name only, and every `aria-label` is preserved verbatim in `SOCIAL_LINKS`.
- `security.yml:24-27` runs `bun i` + `bun audit` against the live advisory DB; this PR touches no `package.json`, no `bun.lock` and no override.

Most likely both are stale runs against a merge ref that can no longer be computed — see the conflict below.

### The marketing-footer hunk no longer applies

`main` restructured that file. The PR's pre-image expects `import Image from 'next/image'` in the import block, `socialMediaIcons` at ~line 52, and the socials in a standalone `<div className="flex flex-row flex-wrap gap-6">` at ~line 211.

On `main`: there is **no** `next/image` import in that file, `socialMediaIcons` is at lines 27-47, and the socials live at ~139-145 inside `<div className="flex flex-wrap items-center gap-8 md:justify-end">` — **shared with the language switcher**. All three hunks fail.

That last detail matters beyond the rebase: `SocialLinks` hard-codes `flex flex-row flex-wrap gap-6`, which is the `/go` footer's design. Dropping it into `main`'s bottom bar nests a `gap-6` row inside a `gap-8` row with no `className` passed to compensate, so the marketing footer's spacing shifts. Either pass the existing classes through or make the wrapper class-free and let callers own layout.

The `/go` footer hunks still match the tree exactly.

### Two more things

1. **The cut-over is incomplete.** `apps/site/src/app/not-found.tsx:10-41` still declares its own copy of the same four social links and renders them at `:86-93`. A third copy surviving defeats the stated "never duplicate them" goal — a URL change would still need two edits.
2. **Silent behaviour change:** marketing social links move from `<Link href>` (same tab) to `<a target="_blank" rel="noopener noreferrer">`. `rel` is correct, but the accessible names (`Visit our Instagram page`) do not announce the new tab. Either keep same-tab behaviour behind a prop, or append "(opens in a new tab)".

### Worth keeping

The `/go` Discord href had a trailing space (`'https://discord.gg/rFY3kc4deK '`); `SOCIAL_LINKS.discord.href` drops it. Genuine bug fix — preserve it through the rebase.

PR #1019, opened 2026-07-12
feat(site): on-demand revalidation webhook for Sanity content

DhruvSingh0905:
## Summary

Content edits currently wait out ISR windows (60-minute listings, 24-hour sitemaps) before appearing on the live site. This PR adds `POST /api/revalidate` so a Sanity webhook can purge the route cache the moment a document is published, unpublished, or deleted — publish in Studio, live in seconds.

## How it works

- Sanity webhook calls `POST /api/revalidate` with an `x-revalidate-secret` header
- Route validates the secret against `SANITY_REVALIDATE_SECRET` (503 if unconfigured, 401 on mismatch)
- On success, purges the full route cache via `revalidatePath('/', 'layout')` — content changes are infrequent enough that a full purge is simpler and safer than per-path mapping
- Logs the triggering document type; tolerates missing/invalid body

## Deploy steps (after merge)

1. Set `SANITY_REVALIDATE_SECRET` in Vercel env (any long random string)
2. In [sanity.io/manage](https://www.sanity.io/manage) → project `website` → API → Webhooks: create a webhook on dataset `production` for create/update/delete, URL `https://toddagriscience.com/api/revalidate`, with HTTP header `x-revalidate-secret: <same value>`

## Testing

- 4 unit tests: unconfigured (503), missing/wrong secret (401), success + full-purge call, missing body
- `bun validate` passes end-to-end

### DIFF
--- a/apps/site/env.example
+++ b/apps/site/env.example
@@ -83,6 +83,11 @@ NEXT_PUBLIC_STRIPE_PUBLISHABLE_KEY="pk_test_your_key_here"
 # Stripe webhook signing secret for /api/stripe/webhook. Private.
 STRIPE_WEBHOOK_SECRET="whsec_your_webhook_secret_here"
 
+# Shared secret for the Sanity on-demand revalidation webhook (/api/revalidate). Private.
+# Generate with `openssl rand -hex 32`; the same value goes in the Sanity webhook's
+# x-revalidate-secret header (sanity.io/manage -> API -> Webhooks).
+SANITY_REVALIDATE_SECRET="your-random-secret-here"
+
 # SMTP settings for internal seed-order notifications. Optional until email delivery is configured.
 ORDER_EMAIL_HOST="smtp.example.com"
 ORDER_EMAIL_PORT="587"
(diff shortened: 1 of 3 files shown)


Jose-Gael-Cruz-Lopez:
this looks good, nice and tight 👍

checked the proxy matcher excludes /api so the webhook actually reaches the route — that's an easy one to miss and it's handled. fail-closed logic is right too: 503 when the secret isn't set, 401 on mismatch, and you're not logging the secret anywhere. tests cover the cases well. also confirmed it doesn't step on #1010 — different files, and the two approaches are complementary (that one shrinks the ISR window, this does the on-demand purge).

few small things, none blocking:

- SANITY_REVALIDATE_SECRET isn't in env.example — can you add it? otherwise nobody knows to set it and the endpoint just 503s silently. this is the one i'd actually want in before merge.
- the secret compare is `provided !== secret` which isn't constant-time. super low risk for a webhook (worst case someone can force a cache purge) but crypto.timingSafeEqual is the airtight version if you want it.
- the success log uses logger.warn, which is dev-gated, so you won't see revalidations firing in prod. the 503 path correctly uses logger.error. if you care about confirming the webhook fires in prod you'd want an always-on level there.

CI's green. nice work

akeight:
Looks great! 

As Jose mentioned, just update the `env.example` with `SANITY_REVALIDATE_SECRET`. You'll also need to sync with Vincent on it to update Vercel and the other devs.

vincenttodd (approved):

PR #1020, opened 2026-07-14
fix(site): repair LinkedIn share URL + add UTM to share buttons

DhruvSingh0905:
## Summary

The article-page **Share** buttons (X / LinkedIn / Facebook / email) had a broken LinkedIn link and no attribution tracking. This fixes both.

## What changed

- **LinkedIn button was broken** — it pointed at `linkedin.com/feed?shareActive&mini=true`, which is not a valid share endpoint and did not reliably open a prefilled share dialog. Switched to the documented `linkedin.com/sharing/share-offsite/?url=` endpoint.
- **Added UTM tags** — every share link (X, LinkedIn, email) now carries `utm_source` / `utm_medium=social` / `utm_campaign=article-share`, so social referrals from these buttons are attributable in analytics (feeds the metrics baseline work).

## Testing

- New unit test covers all four buttons: correct endpoints, UTM tags present, and Facebook clipboard-copy. Mocks `window.open` and asserts the exact URL each button opens.
- `bun validate` passes (format, type-check, lint, tests, build).

## Scope

Touches only `share-article.tsx` + its new test. No behavior change to Facebook (still copy-to-clipboard) or the button layout.

### DIFF
--- a/apps/site/src/components/common/share-article/share-article.test.tsx
+++ b/apps/site/src/components/common/share-article/share-article.test.tsx
@@ -1,0 +1,92 @@
+// Copyright © Todd Agriscience, Inc. All rights reserved.
+
+import { TooltipProvider } from '@/components/ui/tooltip';
+import { fireEvent, render, screen } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
+import type { ReactElement } from 'react';
+import { beforeEach, describe, expect, it, vi } from 'vitest';
+import ShareArticleButtons from './share-article';
+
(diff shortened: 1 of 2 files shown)


rcjasub (approved):
LinkedIn fix is correct (uses the documented sharing/share-offsite/?url= endpoint), UTM tagging is safe (single-encoded, no injection risk), and no regressions — Facebook/layout untouched, tests (4/4) and tsc --noEmit pass locally.

Follow-up (non-blocking): Facebook's button still has no UTM tag, so its referrals won't be attributable like the other three.

Issue #970, opened 2026-06-29
Move farm name and contact to left menu and remove header


vincenttodd:
### Problem Statement

Move the dynamic farm name to the left menu, and remove the header. Also add the contact link to the menu below with the 3 other links.

PR #1012, opened 2026-07-10
Fix/970 move farm contact to left menu

yjm7vj:
## Description
Fixes #970

Moved farm name and primary contact details into the account side menu, removed the old account header from the account shell, and fixed /account routing so the account landing page is protected like the nested account pages.

Changes by file:
**apps/site/src/app/(authenticated)/(shell)/(accounts)/account/(with-shell)/layout.tsx**
Removed the account header from the account layout.
Passed farm/contact shell data into AccountSideMenu.
Updated the loading skeleton to match the headerless layout.

**apps/site/src/app/(authenticated)/(shell)/(accounts)/account/components/account-side-menu/account-side-menu.tsx**
Added farm name and contact details to the top of the account side menu.
Updated the component to accept typed props.

**apps/site/src/app/(authenticated)/(shell)/(accounts)/account/components/account-side-menu/types.ts**
Added AccountSideMenuProps.

**apps/site/src/app/(authenticated)/(shell)/(accounts)/account/components/account-side-menu/account-side-menu.test.tsx**
Updated tests to pass the new side menu props.
Added coverage for rendered farm/contact details.

**apps/site/src/app/(authenticated)/(shell)/(accounts)/account/db.ts**
Extended getAccountShellData() to return farm name plus contact name, email, and phone.

**apps/site/src/proxy/auth.ts**
Added /account as an exact protected route so the account landing page works correctly.

**apps/site/src/proxy/auth.test.ts**
Updated route protection expectations for /account.

**packages/db/scripts/seed-local-db.ts**
Made OpenAI embedding import lazy so local DB seeding can run without OPENAI_EMBEDDINGS_KEY.

## Checklist

- [x] You've included unit or integration tests for your change, where applicable.
- [ ] You've included inline docs for your change, where applicable.
- [ ] Any components that you've modified are accessible.
- [ ] You've used [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/) where appropriate

PR #1021, opened 2026-07-15
Ticket #970 fix (2)

yjm7vj:
## Description
Updates the account routing so account pages no longer render inside the main authenticated app shell. This hides the large platform sidebar on account pages while preserving the expected account page layout shown in the reference screenshot.
This also corrects the earlier scope confusion: the farm/contact info added to the account side menu was already handled in a separate PR, so this fix does not add or modify that account menu content. The account pages continue to show the TODD wordmark, Home link, and farm name as part of the expected account layout.

## Checklist


vincenttodd:
like the strategy, but looks like it didn't stick, at least from what I can tell, somebody more technically should probably review rq

Screenshot 2026-07-15 at 11 12 55 AM

rcjasub (commented):
Overall: The route-group restructure itself is solid — moving (accounts) out of (shell) correctly hides the platform sidebar while account/(with-shell)/layout.tsx preserves the header/side-menu chrome. Import paths and renames all check out cleanly. Three things to fix before merging:

Blocking: onboarding gate is silently dropped for /account/*
requirePlatformOnboardingComplete() (redirects un-onboarded users to /apply) only lives in (shell)/layout.tsx. Since account pages are no longer nested under (shell), and nothing in the new (with-shell) layout or proxy/auth.ts replaces it, a signed-in user who hasn't finished onboarding can now hit /account, /account/security, etc. directly instead of being redirected. Worth adding the check back (or an equivalent) to the account layout, plus a test.

Out of scope: unrelated commit included
This branch carries 7c8821e ("Fix missing OpenAI embeddings key handling"), touching get-embedding.ts/seed-local-db.ts. It's unrelated to account routing, already exists independently on fix/1004-inference-search-error, and isn't on main — looks like a stray merge/rebase artifact. Recommend dropping it from this PR.

Branch is stale
4 commits behind main, including #1006 ("hide account nav link until onboarding is complete") — same surface this PR touches. Please rebase; this also makes the onboarding-gate fix above more urgent, since #1006 assumed the route itself was still server-side protected.

Jose-Gael-Cruz-Lopez:
## This appears to be a duplicate of #1012 that does not implement ticket #970

Both PRs are by the same author, both claim ticket #970 ("move farm contact to the left menu"), and on every substantive line `#1021 ⊂ #1012 ∪ {an obsolete rename}`. The intersection excludes the actual #970 fix.

### It contains no farm/contact change

The ticket asks for farm name and contact details to move into the left menu. In this diff:

- `account-side-menu.tsx` appears only as `similarity index 100%` / `rename` — **zero content change**
- `account/(with-shell)/layout.tsx` — likewise a pure rename
- There is no contact block anywhere in the diff

The only non-rename content edit is in `account/db.ts`:

```diff
-): Promise<{ farmName: string }> {
+): Promise<{
+  farmName: string;
+}> {
```

A whitespace reformat of the same type.

### The rename it performs is already done on main, under a different name

This PR renames `(authenticated)/(shell)/(accounts)/account/**` → `(authenticated)/(accounts)/account/**` (22 renames). `main` consolidated that tree to `(authenticated)/(no-sidebar)/account/**` in #1005. So all 22 land as delete/rename conflicts, and `sidebar.tsx:13` / `sidebar.test.tsx:19` on `main` already point at the *new* path.

### Four files are already on main verbatim

These hunks are identical in #1012 and #1021 and already merged, so they are pure conflict surface with no remaining value:

| File | Already on main |
| --- | --- |
| `apps/site/src/proxy/auth.ts` | `:8-14` (`/account` in `protectedUrls`) and `:137-139` (the exact-match branch) |
| `apps/site/src/proxy/auth.test.ts` | `:62-64` ("should return true for /account without trailing path") |
| `packages/db/scripts/seed-local-db.ts` | `:154` (dynamic `await import('../src/utils/get-embedding')`) |

### The `get-embedding.ts` hunk would break CI

The fourth shared file re-adds a module-scope `throw` that `main` deliberately replaced with a lazy `embeddingsClient()` in #1065. It fires at **import**, and `migrate-remote-db.ts:21` → `importer-lib.ts:12` puts it on the import path of both `pr-database-compatibility-check.yml:94` and `release-database-compatibility-check.yml:92` — neither of which carries `OPENAI_EMBEDDINGS_KEY` (it appears once across all workflows, at `sync-imps-to-staging.yml:19`).

Full detail in the comment on #1007.

### Recommendation

Close as a duplicate of #1012, which does contain the actual farm/contact move. Nothing here needs salvaging into #1012 — the rename it wants is already on `main` under `(no-sidebar)/`.

For #1012 itself, the remaining work is: rebase onto `(no-sidebar)/account/`, drop these four tail files, delete the now-orphaned `account-header.tsx`, and restore an `<h1>` — removing `AccountHeader` currently deletes the only `<h1>` and the only `role="banner"` landmark in the entire account subtree (`account-header.tsx:16,34`).

PR #1022, opened 2026-07-16
fix: pin third-party GitHub Actions to immutable commit SHAs 

### LINKED ISSUE #1015: ci: pin GitHub Actions to commit SHAs instead of version tags
## Problem Statement

Third-party GitHub Actions in `.github/workflows/` are pinned to mutable version tags (e.g. `@v5`) rather than immutable commit SHAs. A tag can be repointed by the action's owner — or by an attacker who compromises that account — to code the maintainers never reviewed. Because the action then runs inside our CI with whatever `GITHUB_TOKEN` scope the workflow grants, a hijacked tag is a direct supply-chain path into the repo.

## Context

Surfaced during review of #1011 (repo cleanup). It is **pre-existing** — that PR touches no workflow files — and was scoped as a post-merge hygiene item, not a blocker.

**Third-party actions** currently tag-pinned (highest priority):

- `codecov/codecov-action@v5`
- `dependabot/fetch-metadata@v2`
- `lewagon/wait-on-check-action@v1.8.0`
- `oven-sh/setup-bun@v2`

**GitHub-official actions** also tag-pinned (lower risk — GitHub controls the org — but the same class, and GitHub's own hardening guide recommends SHA-pinning these too):

- `actions/checkout@v6`
- `actions/setup-node@v6`
- `actions/upload-artifact@v7`

This pairs with the sibling issue on missing workflow `permissions:` blocks: SHA-pinning limits *what code runs*, least-privilege limits *what that code can do*. Both together give defense in depth.

## Constraints

- Keep a readable version reference. Pin to the full 40-character commit SHA with a trailing comment of the human tag, e.g. `oven-sh/setup-bun@<sha> # v2.0.2`.
- Do not silently bump versions while pinning — pin to the SHA of the tag currently in use, then let Dependabot propose upgrades.

## Technical Details

- Replace each `uses: owner/action@vX` with `uses: owner/action@<40-char-sha>  # vX`, resolving the SHA the tag currently points to.
- Configure Dependabot for the `github-actions` ecosystem so it opens PRs to advance the pinned SHAs as new releases land (keeps pins current without reintroducing mutable tags).
- Prioritize the four third-party actions first; the `actions/*` official ones can follow.
- References:
  - https://docs.github.com/actions/security-guides/security-hardening-for-github-actions#using-third-party-actions
  - https://docs.github.com/code-security/dependabot/working-with-dependabot/keeping-your-actions-up-to-date-with-dependabot

## Use Case

As a maintainer, I want every action pinned to an immutable commit SHA, so that a repointed or compromised tag on a third-party action cannot inject unreviewed code into our CI runs.

StoneTimber:
## Description

Pinned all third-party and GitHub-official Actions in .github/workflows/ from mutable version tags (e.g. @v5) to immutable commit SHAs (e.g. @abc123... # v5). This prevents a repointed or compromised tag from injecting unreviewed code into CI runs. No logic or behavior was changed — this is a security hygiene update only.

Fixes #1015 

## Checklist

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -13,15 +13,15 @@ jobs:
 
     steps:
       - name: Checkout code
-        uses: actions/checkout@v6
+        uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6
 
       - name: Setup Node.js
-        uses: actions/setup-node@v6
+        uses: actions/setup-node@249970729cb0ef3589644e2896645e5dc5ba9c38 #v6
         with:
           node-version: '20'
 
       - name: Setup Bun
-        uses: oven-sh/setup-bun@v2
+        uses: oven-sh/setup-bun@0c5077e51419868618aeaa5fe8019c62421857d6 #v2
         with:
           bun-version: 'latest'
 
@@ -38,7 +38,7 @@ jobs:
         run: bun run build-storybook
 
       - name: Upload Next.js build artifacts
-        uses: actions/upload-artifact@v7
+        uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a #v7
         if: always()
         with:
           name: nextjs-build
@@ -48,7 +48,7 @@ jobs:
           retention-days: 7
 
       - name: Upload Storybook build artifacts
-        uses: actions/upload-artifact@v7
+        uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a #v7
         if: always()
         with:
           name: storybook-build
--- a/.github/workflows/code-quality.yml
+++ b/.github/workflows/code-quality.yml
@@ -13,15 +13,15 @@ jobs:
 
     steps:
       - name: Checkout code
-        uses: actions/checkout@v6
+        uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 #v6
 
       - name: Setup Node.js
-        uses: actions/setup-node@v6
+        uses: actions/setup-node@249970729cb0ef3589644e2896645e5dc5ba9c38 #v6
         with:
           node-version: '20'
 
       - name: Setup Bun
-        uses: oven-sh/setup-bun@v2
+        uses: oven-sh/setup-bun@0c5077e51419868618aeaa5fe8019c62421857d6 #v2
         with:
           bun-version: 'latest'
 
--- a/.github/workflows/db-reconcile.yml
+++ b/.github/workflows/db-reconcile.yml
@@ -53,7 +53,7 @@ jobs:
       STAGING_DATABASE_PEM_CERT: ${{ secrets.STAGING_DATABASE_PEM_CERT || secrets.DATABASE_PEM_CERT }}
     steps:
       - name: Checkout code
-        uses: actions/checkout@v6
+        uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 #v6
 
       - name: Install PostgreSQL 18 client
         run: |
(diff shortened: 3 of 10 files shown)


Jose-Gael-Cruz-Lopez:
## Right idea — but it skips the one workflow that matters most, and silently downgrades one action

Pinning to immutable SHAs is exactly right and this should land early: it is low blast radius and every other PR benefits from hardened CI. Three things to fix first.

### 1. `sync-imps-to-staging.yml` is not in this PR — and it is the highest-value target in the repo

This PR touches 10 files. `.github/workflows/` contains **11**. The omitted one:

| File:line | Action | Still on a mutable tag |
| --- | --- | --- |
| `sync-imps-to-staging.yml:29` | `actions/checkout@v6` | yes |
| `sync-imps-to-staging.yml:31` | `google-github-actions/auth@v3` | **yes — and this is its only occurrence in the repo, so it is pinned nowhere** |
| `sync-imps-to-staging.yml:37` | `oven-sh/setup-bun@v2` | yes |

That workflow's `env:` block exposes `OPENAI_EMBEDDINGS_KEY`, `STAGING_DATABASE_PASSWORD`, `STAGING_DATABASE_DATABASE`, `DATABASE_PEM_CERT` and `STAGING_DATABASE_PEM_CERT`, and `google-github-actions/auth@v3` performs GCP **Workload Identity Federation** against `projects/609771777761/…/nightcrawler-github` for service account `nightcrawler-sheets@…`.

A hijacked `v3` tag there mints live GCP credentials **inside a job already holding the staging DB password and PEM cert**. Every other third-party action in the repo is covered by this PR — this is the gap.

### 2. It downgrades `lewagon/wait-on-check-action` by a minor version

```diff
-        uses: lewagon/wait-on-check-action@v1.8.0
+        uses: lewagon/wait-on-check-action@eb1f8d8fb8af5d2d3c1978a3fae7ef38eaa587e8 # v1.8.0
```

But `.github/workflows/dependabot-auto-approve.yml:46` on `main` is already `@v1.9.0`. Merging as-is rolls it back — and this is also the hunk causing the conflict. Re-pin to the `v1.9.0` SHA.

### 3. SHA↔tag correspondence is unverifiable by review, with no gate

Every pin is a bare 40-char SHA plus a hand-written comment. Whether `df4cb1c069e1874edd31b4311f1884172cec0e10` really is `actions/checkout@v6` is precisely the property this PR exists to guarantee, and a wrong or malicious SHA would be invisible to a reviewer. Worth making mechanical:

```bash
gh api repos/actions/checkout/git/ref/tags/v6 --jq .object.sha
```

…or adding `zizmor` / `pin-github-action` to `code-quality.yml` as a standing gate.

### Minor

- Comment style is inconsistent **within a single file** — `build.yml` uses `# v6` on checkout but `#v6` on setup-node and setup-bun. Dependabot's `github-actions` updater parses that trailing comment to know what to bump, so normalise to `# vX.Y.Z`.
- Annotating with a moving major (`# v6`) rather than the exact release loses the information needed to tell whether a pin is stale.

### Good news you do not need to solve

`.github/dependabot.yml:89-101` already enables the `github-actions` ecosystem with a `patterns: ['*']` group on a monthly schedule, so pinned SHAs will be kept fresh automatically. Pinning here does not create stale-action rot.

Jose-Gael-Cruz-Lopez:
Closing: superseded by #1084, which merged. Every action across all 11 workflows is now SHA-pinned on `main` — I verified zero mutable tags remain.

One reason to prefer #1084's version rather than rebasing this one: `actions/checkout` here is pinned to `df4cb1c0`, but `v6` currently resolves to `d23441a4` (confirmed against the GitHub API). Merging this would have pinned a stale commit under a `# v6` comment that looked correct. #1084's SHAs all verify — including `lewagon/wait-on-check-action@v1.9.0`, which is an annotated tag and needs dereferencing to get the commit rather than the tag object.

Thanks for starting this — #1015 is closed by the merge.

PR #1028, opened 2026-07-22
chore(deps): fix high security audit findings

longlong9889:
## Description

Fixes #{1027}

Problem Statement

The security audit on main is failing because several transitive dependencies have high-severity advisories. This blocks the security workflow and leaves the repository reporting known vulnerable package versions.

Context

This PR updates dependency overrides and the Bun lockfile to resolve the high-severity audit findings reported by bun audit--audit-level=high.

The affected packages were:
  fast-uri
  linkify-it
  sharp

 The fix bumps those packages to patched versions through root-level overrides and refreshes bun.lock so the resolved dependency graph no longer includes the vulnerable versions.

Validation completed:

  bun audit --audit-level=high
  bun run format:check

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -226,11 +226,11 @@
     "brace-expansion": "^5.0.7",
     "dompurify": "3.3.2",
     "esbuild": "^0.28.1",
-    "fast-uri": "^3.1.2",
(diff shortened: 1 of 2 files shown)


rcjasub (approved):
LGTM

PR #1029, opened 2026-07-23
fix(site): correct competencies scroll animation

StoneTimber:
## Description

1. Fixes the scroll-driven animation in the About page's competencies section 
2. Adds the Pillars animation in between the Competencies section and Partners section.

## Checklist


apps/site/src/app/(unauthenticated)/[locale]/(marketing)/about/components/competencies-section/competencies-section.tsx
```diff
@@ -166,64 +205,67 @@ function CompetencyCircle({
...
         position: 'absolute',
-        top,
+        top: '50%',
         x,
         y,
       }}
       className="flex items-center justify-center text-center left-1/2"
     >
-      <div className="flex size-48 md:size-64 flex-col items-center justify-center rounded-full border-0 bg-[#AB844F]/20 p-6 md:p-8 transition-[background-color] duration-500 ease-[cubic-bezier(0.25,0.1,0.25,1)] hover:bg-black/10">
+      <div className="relative flex size-48 md:size-64 flex-col items-center justify-center rounded-full border-0 bg-[#AB844F]/20 p-6 md:p-8 transition-[background-color] duration-500 ease-[cubic-bezier(0.25,0.1,0.25,1)] hover:bg-black/10">
```
Comment (akeight) [inline apps/site/src/app/(unauthenticated)/[locale]/(marketing)/about/components/competencies-section/competencies-section.tsx]: Let's remove the `hover:bg-black/10` here so the colors stay consistent.

apps/site/src/app/(unauthenticated)/[locale]/(marketing)/about/components/pillars-section/pillars-section.stories.tsx
```diff
@@ -0,0 +1,42 @@
+// Copyright © Todd Agriscience, Inc. All rights reserved.
+
+import type { Meta, StoryObj } from '@storybook/nextjs-vite';
+import PillarsSection from './pillars-section';
+
+const placeholderCopy: Record<string, string> = {
+  'pillars.title': 'Pillars',
+  'pillars.subtitle':
+    'Placeholder subtitle describing the five pillars that guide our work with farms.',
```
Comment (akeight) [inline apps/site/src/app/(unauthenticated)/[locale]/(marketing)/about/components/pillars-section/pillars-section.stories.tsx]: Is there a subtitle and description available for these? Sync with Vincent on that if you haven't already. :)

akeight (commented):
Overall, really great work! Tested for desktop, mobile, and responsive screen sizes in dev mode. All look clean. Let's make sure to verify in Vercel preview mode- the last competencies component got really glitchy in prod.

Here are a few remarks from senior frontend expert Opus. If you can to implement these, go ahead. Just verify there's no regression when added.

### Accessibility — high priority
1. **No prefers-reduced-motion (blocking)**
Global CSS only kills CSS animation/transition:
in globals.css, lines 112-121:
`@media (prefers-reduced-motion: reduce) {`
  `...`
  `* {`
   ` animation-duration: 0.01ms !important;`
    `...`
 ` }`
`}`

These sections drive motion via Framer Motion JS (useScroll, stickyY, infinite rotate: 360). That is unaffected. Users who opted out of motion still get:
- 400vh + 500vh scroll “sticky” scrubbing
- Mobile pillars infinite 40s spin
_Use useReducedMotion() and fall back to the static mobile-style layout (or static desktop content, no scroll scrubbing / no infinite rotate). WCAG 2.3.3 / 2.2.2 territory._

2. **Return null until width is known (blocking for SSR / a11y / SEO)**
competencies-section.tsx, lines 31-33:
 `if (windowWidth === undefined) {`
    `return null;`
 ` }`

Same pattern in pillars. Consequences:
- No content in SSR HTML — crawlers and no-JS users miss competencies + pillars entirely.
- CLS when content pops in after hydration.
- Screen reader users get a late content insert.
_Prefer: always render the static/mobile content shell (semantics + copy), then enhance to scroll-driven desktop after mount / matchMedia. Don’t withhold content to protect an animation._

### Performance
1. **Shared hook tax: useWindowWidth is buggy**
useWindowWidth.tsx, lines 11-24:
  `useEffect(() => {`
    `const handleResize = () =>`
      `setTimeout(() => setWidth(window.innerWidth), 100);`
    `window.addEventListener('resize', handleResize);`
    `...`
  `});` // ← no dependency array

Effect re-binds on every render.
setTimeout is never cleared (leak / stale updates).
Claimed “throttled once per 100ms” isn’t real throttling.
_Both sections pay this cost. Fix the hook ([] deps, matchMedia('(min-width: 1024px)', clear timeouts) — better than measuring full width for a binary layout choice._

vincenttodd (approved):
Looks great!

PR #1034, opened 2026-07-29
feat: resolve supabase auth redirects to vercel preview urls

### LINKED ISSUE #497: Implement Vercel Preview URLs for Supabase
### Problem Statement

Currently, we only redirect Supabase URLs on either `localhost:3000` or `toddagriscience.com`, which can lead to poor DX when testing in Vercel preview environments. We can change this with a few environment variables though - the following logic from [Supabase](https://supabase.com/docs/guides/auth/redirect-urls#vercel-preview-urls) should be implemented in the codebase. Update any environment variables that need to be updated in `env.example`.

### Context

_No response_

### Constraints

This should not change any existing Supabase logic. It should only add this functionality and refactor any code that is related to this change.

### Technical Details

_No response_

### User Story

- As a developer, I want to be able to test my changes on Vercel Preview deployments without having to manually change the URL

### Use Case

_No response_

Orlando275:
## Description

Fixes #497

Supabase auth emails always resolved their redirect from `NEXT_PUBLIC_BASE_URL`, so anyone testing a Vercel preview deployment was bounced to production instead of landing back on the deployment they were testing.

### What changed

Adds `getAuthRedirectBaseUrl()` to `apps/site/src/lib/env.ts`:

- On `preview` deployments it prefers `VERCEL_BRANCH_URL` — stable across redeploys of the same branch, so a single Supabase redirect entry covers every push to that branch — and falls back to the per-deployment `VERCEL_URL`.
- Production and local development keep resolving to `NEXT_PUBLIC_BASE_URL`, byte for byte as before.
- Reads both the `NEXT_PUBLIC_VERCEL_*` and unprefixed `VERCEL_*` variants, matching the existing pattern in `lib/logger.ts`.
- `normalizeOrigin()` adds `https://` (Vercel exposes hosts without a scheme) and strips trailing slashes.

The four call sites that assembled the origin by hand now go through the helper:

| File | Before |
|---|---|
| `lib/auth-server.ts` `signUpUser` | `NEXT_PUBLIC_BASE_URL?.replace(...) ?? '…'` |
| `lib/auth-server.ts` `inviteUser` | `NEXT_PUBLIC_BASE_URL + '/auth/accept-invite'` |
| `lib/auth-server.ts` `resendEmailInvite` | same concatenation |
| `signup/actions.ts` | its own fallback chain |

That last one fell back to `NEXT_PUBLIC_PRODUCTION_DOMAIN`, which carries no scheme, and handed it to `buildSignupUrl()`. The helper normalizes it.

`env.example` documents the three Vercel variables, and notes that the preview origin has to be allow-listed under Supabase → Authentication → URL Configuration → Redirect URLs.

### Note for the reviewer

The global Vitest mock in `vitest.setup.intl.js` replaced the whole `env` module, so any newly exported symbol resolved to `undefined` in all 117 test files. It is now a partial mock (`importActual()` + spread) that still stubs the `env.baseUrl` the metadata tests depend on.

### Out of scope

`resetPasswordForEmail` in `lib/actions/auth.ts` passes no `redirectTo` and relies on the Supabase project Site URL. Giving it one would add new Supabase behavior rather than re-source an existing redirect, which the issue's constraint rules out.

### Verification

`CI=true TURBO_ENV_MODE=loose LC_ALL=en_US.UTF-8 bun validate` passes end to end: format:check, lint (0 errors), type-check, 565 site tests across 117 files + 11 db tests, build.

Twelve new cases in `lib/env.test.ts` cover the precedence chain, production and localhost staying untouched, trailing slashes, the schemeless fallback, the unprefixed variables, and a declared-but-blank `NEXT_PUBLIC_VERCEL_ENV`. Two integration cases in `lib/auth.test.ts` assert that under `VERCEL_ENV=preview` the `emailRedirectTo` of `signUpUser` and the `redirectTo` of `inviteUser` point at the preview host rather than `toddagriscience.com` — both fail if the production code is reverted.

## Checklist


Orlando275:
Both review comments applied in 1366a05.

**Whitespace-only env values (medium).** Confirmed and fixed. Every variable now goes through a `readEnv()` helper that trims and maps blank values to `undefined`, so a stray space falls through to the next candidate in the chain instead of short-circuiting it. `normalizeOrigin()` can no longer be reached with a blank value, so the malformed `https://` origin is gone.

Four cases lock it in, all verified to fail against the previous implementation:

- `treats a whitespace-only base URL as unconfigured`
- `never returns a scheme-only origin when every variable is blank` — asserts `new URL('/signup', result)` does not throw, which is the `buildSignupUrl` crash path
- `skips a whitespace-only branch URL in favour of the deployment URL`
- `falls back to the base URL when every Vercel host is blank`

**Leaked `NEXT_PUBLIC_BASE_URL` (low).** Confirmed and fixed. The preview `describe` block now captures all three keys it touches in `beforeEach` and restores them in `afterEach`, deleting the ones that were previously unset rather than leaving a value behind.

`CI=true TURBO_ENV_MODE=loose LC_ALL=en_US.UTF-8 bun validate` is green: 574 site tests across 118 files, plus 11 db tests.

rcjasub (approved):
LGTM
Verified the implementation against the description rather than taking it at face value:

- getAuthRedirectBaseUrl() in lib/env.ts correctly implements the preview → VERCEL_BRANCH_URL → VERCEL_URL → prod fallback chain, reading both prefixed/unprefixed vars (consistent with the existing pattern in lib/logger.ts).
- All 4 call sites (signUpUser, inviteUser, resendEmailInvite, resendApprovedApplicantActivationEmail) now go through the shared helper — nice dedupe of the previously hand-rolled URL concatenation.
- readEnv/normalizeOrigin correctly handle blank/whitespace env vars, schemeless hosts, and trailing slashes.
- resetPasswordForEmail was correctly left alone per the stated scope boundary.
- vitest.setup.intl.js partial-mock fix (importActual() + spread) is the right call — avoids breaking the other 116 test files while still letting the new helper read real process.env.
- Production/local behavior is byte-for-byte unchanged — no regression risk there.

Ran it myself to confirm:
- bun vitest run → 574/574 tests passing (118 files)
- tsc --noEmit → clean
- eslint . → 0 errors (2 pre-existing, unrelated warnings)

PR #1035, opened 2026-07-31
feat(site): replace account date inputs with shadcn calendar

### LINKED ISSUE #748: Swap calendar for ShadCN calendar
### Problem Statement

This issue should edit the ui in the `/account` pages to use the [ShadCN style calender](https://ui.shadcn.com/docs/components/radix/calendar) instead of the one currently in place for style editing.

### Context

_No response_

### Constraints

_No response_

### Technical Details

_No response_

### User Story

_No response_

### Use Case

_No response_

Orlando275:
## Description

Fixes #748

The two date fields on `/account/management-zones/[zone]` — `Rotation year` and `NPK last used` — used native `<input type="date">`. They now open the shadcn calendar inside a popover. Since the trigger is a button rather than an input, they are wired through react-hook-form's `Controller` instead of `register`.

These were the only native date pickers under `/account`; the other two account route groups (`/account/agreement` and `/account/reset-password`) have no date UI.

### Notes for review

- **Timezone.** The `date` columns round-trip at UTC midnight, while react-day-picker builds its dates at local midnight. Without converting on both ends, the stored day shifts by one for users east of UTC. `toCalendarDate` / `fromCalendarDate` handle this, and a test pins the submitted value to `2026-07-15T00:00:00.000Z`.

- **Year range.** With `captionLayout="dropdown"`, react-day-picker caps navigation at the end of the current year by default, which would make future rotation dates unreachable — a regression against the native input. The range is declared explicitly as 50 years back and 10 forward. Happy to change the window if you'd prefer different bounds; it's two constants.

- **Tailwind v4.** The shadcn CLI serves the Tailwind v3 flavour of `calendar.tsx`, because `components.json` declares a `tailwind.config.ts`. Its `h-[--cell-size]` classes compile to invalid CSS under v4 (`height: --cell-size`, which browsers drop), which would have left the calendar misaligned. They were converted to the v4 `h-(--cell-size)` syntax and verified against the compiled production CSS. Note that `type-check`, `lint`, `test` and `build` all pass either way, since none of them evaluate CSS.

- **Accessibility.** A `<label for>` does not name a `<button>`, so each trigger carries an `aria-label` combining its label and current value.

- `calendar.tsx` and `popover.tsx` are shadcn output kept as generated, apart from the Tailwind fix and the copyright header. They carry no stories or tests, matching the other 22 components in `src/components/ui/`.

## Checklist


Jose-Gael-Cruz-Lopez:
Closing in favour of #1087, which rebases this work onto `main` with three fixes folded in. Verified against `main` (5dc3f272):

- `popover.tsx:24` here uses Tailwind **v3** syntax `origin-[--radix-popover-content-transform-origin]`. This repo is on `tailwindcss ^4.2.2` (`apps/site/package.json:112`), where a CSS-variable value needs parentheses — so that utility emits nothing and the popover animates from its own centre instead of the trigger.
- No `data-slot` attributes. `calendar.tsx` styles itself with `[[data-slot=popover-content]_&]:bg-transparent`, so without them the calendar keeps `bg-background` inside a `bg-popover` panel.
- `form.tsx:102-103` keeps `<Label htmlFor={name}>` pointing at `<Button id={name}>`. A `<label for>` cannot name a button, so the association is inert and clicking the label does nothing — a regression against the native input it replaces.

#1087 fixes all three and additionally uses `outline-hidden` (v4), matching main's own convention at `components/ui/dialog.tsx:74`.

Worth noting your branch carried a **correct `bun.lock`**; #1087's rebase dropped it, so that has to be regenerated there. The date-picker design itself is sound and is what's being kept.

PR #1053, opened 2026-08-03
fix(site): keep the footer location label out of the language button

### LINKED ISSUE #1052: [Bug]: Footer location label opens the language menu
### Bug Description

In the marketing footer, the language label and the location label ("Español" / "Estados Unidos", or "English" / "United States") look like two separate controls, but they are a **single button**. Clicking either one opens the same language menu with only two options: English and Español.

Clicking "Estados Unidos" therefore does something the label doesn't promise — it opens a language picker, not a region picker.

The cause is in `apps/site/src/app/(unauthenticated)/[locale]/(marketing)/components/footer/footer.tsx:148-162`, where one `<button>` wraps both spans:

```tsx
<button
  type="button"
  aria-haspopup="menu"
  aria-expanded={langOpen}
  onClick={() => setLangOpen(!langOpen)}
  aria-label="Change language"
>
  <span className="text-[15px] font-normal">{currentLanguageLabel}</span>
  <span className="text-[15px] font-normal text-foreground/50">{t('location')}</span>
</button>
```

Both spans sit inside the button's hit area, so the whole strip is one target wired to `setLangOpen`.

The location text is also purely decorative: `t('location')` resolves to a hardcoded string in `apps/site/src/messages/footer/en.json:4` (`"location": "United States"`) and `apps/site/src/messages/footer/es.json:4` (`"Estados Unidos"`). There is no region state, no region detection and no region switching anywhere in the project — nothing changes if a user were able to "pick" a region.

There is an accessibility dimension too: the button's `aria-label` is `"Change language"`, so screen reader users hear one control named "Change language" while sighted users see what reads as two. And the 50% opacity on the location span is the same treatment the codebase uses for non-interactive text, which is exactly why it reads as a static label rather than part of the button.

### Steps to Reproduce

1. Open `https://toddagriscience.com/` (or any marketing page) and scroll to the footer
2. Click "Español" — a menu appears with English / Español
3. Close it, then click "Estados Unidos"
4. The same English / Español menu appears, instead of anything region-related

### Expected Behavior

Clicking "Estados Unidos" should not open the language menu.

There are two reasonable resolutions and this needs a product/design call before anyone writes code:

- **If the location is meant to be a static label** (most likely — the 50% opacity matches how non-interactive text is styled elsewhere): move it outside the `<button>` into a sibling `<span>` so only the language label is clickable. Small change, contained to `footer.tsx`.
- **If a real region selector is intended**: it needs to be designed and built. That means deciding which regions exist, what selecting one actually changes, and how it relates to locale. Significantly larger, and out of scope for a bug fix.

### Screenshots/Videos

_No response_

### Browser

Edge

### Device Type

Desktop

### Language/Locale

Spanish (es)

### Additional Context

**Not a regression from the recent footer redesign.** Checking the file at the commit before "Footer redesign" (#1023, `39999b1`) shows the same single-button structure already in place, with `t('location')` inside it. This has been present since the language selector was introduced in #401.

The language switching itself works correctly — `handleLocaleChange` sets the `NEXT_LOCALE` cookie and navigates, and that part is fine. The defect is purely that the location label is inside the button's hit area.

Orlando275:
## Description

Fixes #1052

In the marketing footer, the language label and the location label looked like two separate controls but were a single `<button>` wired to `setLangOpen`. Clicking "Estados Unidos" / "United States" opened the language menu — something its label never promised.

This takes the first of the two resolutions listed in the issue: the location is a static label, so it moves out of the button.

### What changed

`footer.tsx` — the location `<span>` is now a sibling of the button rather than its child:

- The `flex items-center gap-3 md:gap-6` moved from the button to the wrapping `div`, which is now the element with two children. Spacing is visually identical.
- The language label's inner `<span>` is gone; the button has a single text child and already carries `text-[15px] font-normal`.
- The wrapper `div` still contains both elements, so the menu's `absolute right-0` anchors exactly where it did before. No visual change to the dropdown.
- A comment records why the location is not a control and that multi-country support is planned.

`footer.test.tsx` — two tests: the location is not inside a `<button>` and clicking it opens no menu; the language button still opens the menu with English and Español.

### Out of scope

The button keeps `aria-label="Change language"`. The issue notes the accessible name doesn't match the visible text, which is real, but the same `aria-label` is used in `src/components/common/locale-switcher/locale-switcher.tsx` — it's a codebase-wide pattern, and changing it in one place only would create an inconsistency. After this change the label is at least accurate: the button is now solely the language control. Worth a separate issue.

No region selector was built. That was the issue's second resolution and it needs product and design decisions first.

## Checklist

### DIFF
--- a/apps/site/src/app/(unauthenticated)/[locale]/(marketing)/components/footer/footer.test.tsx
+++ b/apps/site/src/app/(unauthenticated)/[locale]/(marketing)/components/footer/footer.test.tsx
@@ -1,6 +1,11 @@
 // Copyright © Todd Agriscience, Inc. All rights reserved.
 
-import { screen, renderWithNextIntl } from '@/test/test-utils';
+import {
+  screen,
+  within,
+  fireEvent,
+  renderWithNextIntl,
+} from '@/test/test-utils';
 import Footer from './footer';
 import '@testing-library/jest-dom';
 import { describe, it, expect } from 'vitest';
@@ -81,6 +86,32 @@ describe('Footer', () => {
     ).toBeInTheDocument();
     expect(
       screen.getByRole('link', { name: 'Visit our YouTube channel' })
+    ).toBeInTheDocument();
+  });
+
+  it('renders the location as a static label outside the language button', async () => {
+    renderWithNextIntl(<Footer />);
+
+    const location = screen.getByText('United States');
+    expect(location.closest('button')).toBeNull();
+
+    await fireEvent.click(location);
+    expect(screen.queryByRole('menu')).not.toBeInTheDocument();
+  });
+
+  it('opens the language menu from the language button', async () => {
+    renderWithNextIntl(<Footer />);
+
+    await fireEvent.click(
+      screen.getByRole('button', { name: 'Change language' })
+    );
+
+    const menu = screen.getByRole('menu');
+    expect(
+      within(menu).getByRole('menuitem', { name: 'English' })
+    ).toBeInTheDocument();
+    expect(
+      within(menu).getByRole('menuitem', { name: 'Español' })
     ).toBeInTheDocument();
   });
 
--- a/apps/site/src/app/(unauthenticated)/[locale]/(marketing)/components/footer/footer.tsx
+++ b/apps/site/src/app/(unauthenticated)/[locale]/(marketing)/components/footer/footer.tsx
@@ -144,22 +144,27 @@ const Footer = () => {
               {val.icon}
             </Link>
           ))}
-          <div className="relative inline-flex">
+          <div className="relative inline-flex items-center gap-3 md:gap-6">
             <button
               type="button"
               aria-haspopup="menu"
               aria-expanded={langOpen}
               onClick={() => setLangOpen(!langOpen)}
-              className="flex flex-row items-center gap-3 text-[15px] font-normal cursor-pointer focus:outline-none focus:ring-0 md:gap-6"
+              className="text-[15px] font-normal cursor-pointer focus:outline-none focus:ring-0"
               aria-label="Change language"
             >
-              <span className="text-[15px] font-normal">
-                {currentLanguageLabel}
-              </span>
-              <span className="text-[15px] font-normal text-foreground/50">
-                {t('location')}
-              </span>
+              {currentLanguageLabel}
             </button>
+
+            {/*
+              Static label, not a control. The site serves a single region and
+              has no region state, detection or switching, so there is nothing
+              to pick yet. Kept visible as a placeholder for the planned
+              multi-country support.
+            */}
+            <span className="text-[15px] font-normal text-foreground/50">
+              {t('location')}
+            </span>
 
             {langOpen && (
               <>


Orlando275:
### The failing `Security Audit` is pre-existing, not from this diff

This PR changes two `.tsx` files in the marketing footer and touches no dependencies, but `Security Audit` is red. Here's the diagnosis so whoever reviews this doesn't have to redo it.

#### What the check actually does

`.github/workflows/security.yml` runs two steps:

```yaml
- run: bun i
- run: bun audit --audit-level=high
```

`bun audit` compares the installed dependency tree against the public advisory database and exits non-zero on anything rated high or above. Since `bun i` installs exactly what `bun.lock` pins, the check always evaluates the lockfile that's committed to the branch. A PR that doesn't touch `bun.lock` cannot change this result in either direction.

#### What it found

```
fast-uri  >=3.0.0 <3.1.5    (transitive, via ajv)
  high: host confusion via backslash authority introducer — GHSA-7p8r-x3mc-p8w7

undici    >=7.0.0 <7.29.0   (direct, plus jsdom / newrelic / sanity)
  high: cross-user information disclosure and parse-time crash — GHSA-4cwx-7wf7-3272

2 vulnerabilities (2 high)
```

#### Why it started failing now

This is the first run of the workflow since Aug 1:

| Run | Branch | Result |
|---|---|---|
| Aug 1, 16:22 | `main` | success |
| Aug 1, 16:18 | `feat/vercel-preview-supabase-urls` | success |
| Aug 1, 13:12 | several dependabot PRs | success |
| Aug 3, 20:53 | `fix/1052-footer-location-label` | **failure** |

Both advisories were published in between. The same lockfile that passed on Aug 1 fails today — the dependency tree didn't change, the advisory database did. `main` is only green in the sense that it hasn't been re-run since; any push to it now fails identically. Dependabot has already flagged the same thing on the default branch.

#### The fix is a lockfile refresh, not a `package.json` edit

The `overrides` block in the root `package.json` already permits the patched releases:

| Package | Override range | Locked in `bun.lock` | First patched |
|---|---|---|---|
| `fast-uri` | `^3.1.3` | `3.1.4` | `3.1.5` |
| `undici` | `^7.28.0` | `7.28.0` | `7.29.0` |

`^3.1.3` allows `3.1.5` and `^7.28.0` allows `7.29.0`, so the ranges are fine — only the resolutions are stale. `bun install` won't move them, because it honours the lockfile. Re-resolving is what's needed:

```bash
bun update undici fast-uri
```

That rewrites `bun.lock` with the newest version each range allows, clearing both advisories. Raising the override floors to `^3.1.5` / `^7.29.0` afterwards would additionally stop the lockfile from regressing and document why the pins exist.

This belongs in its own `fix(deps)` PR rather than here: it rewrites the lockfile across all four workspaces, and bundling it with a footer markup change makes both harder to review. Same shape as #1028.

Since `Security Audit` is a required check, this PR stays blocked until that lands — nothing in the footer change can unblock it.

rcjasub (approved):
The fix is exactly what the PR description claims, isolated to one commit (0e7e64b), and matches the "static label" resolution the issue called for.  LGTM

PR #1054, opened 2026-08-03
Fix 1032 contact redesign

### LINKED ISSUE #1032: Redesign the /contact page
### Problem Statement

This issue will redesign the contact page based on Figma design. 

### Context

_No response_

### Constraints

_No response_

### Technical Details

_No response_

### User Story

_No response_

### Use Case

_No response_

yjm7vj:
## Description

Fixes #1032 
Updates the public `/contact` page to render the Sanity-managed contact form instead of the previous hardcoded support form.

## Checklist


rcjasub (commented):
Nice cleanup reusing the existing DynamicForm/Sanity form system instead of
inventing new form logic — good reduction in surface area, and the server
action already re-validates + honeypots server-side.

One regression before this can merge: `public-inquiry-modal.tsx` (login page's
"Help" modal) links to `/contact?intent=forgot-email` and
`/contact?intent=contact-support`, expecting the contact page to swap in
"Last Known Email Address" copy for the forgot-email case. The new
DynamicForm ignores `intent` entirely, so that context is now silently lost
for anyone using that flow. Can we either wire intent into the Sanity form
config, or drop/update those modal options to match the new page?

Also some now-dead code left behind that should be removed in this PR:
- contact/actions.ts (submitPublicInquiry) + actions.test.ts
- contact/types.ts (publicInquirySchema)
- messages/contact/{en,es}.json — the `contactPage.*` namespace, incl. the
  orphaned `forgotEmail` string

Non-blocking: getFormBySlug swallows Sanity errors and returns null, which
now 404s /contact instead of always rendering (old behavior). Worth
confirming the "contact" form doc is actually published in prod Sanity, and
that a 404 is an acceptable failure mode if Sanity has a hiccup.

rcjasub (approved):
LGTM

Issue #1057, opened 2026-08-04
[Bug]: Switching from English to Spanish 404s — the locale prefix overwrites the first path segment


Orlando275:
### Bug Description

Using the language menu in the marketing footer to switch from English to Spanish 404s on every page except the homepage. Switching back from Spanish to English appears to work, but only because next-intl rescues it with a redirect.

The cause is in `apps/site/src/app/(unauthenticated)/[locale]/(marketing)/components/footer/footer.tsx:80-89`:

```ts
const handleLocaleChange = (newLocale: string) => {
  setLangOpen(false);
  const segments = pathname.split('/');
  segments[1] = newLocale;
  const newPath = segments.join('/') || `/${newLocale}`;
  ...
```

`segments[1] = newLocale` **replaces** the first path segment instead of inserting a locale prefix. That is only correct when a locale prefix is already there.

Routing is configured with `localePrefix: 'as-needed'` and `defaultLocale: 'en'` (`apps/site/src/i18n/config.ts:20-26`), so English pages are served without a prefix and Spanish pages with one:

| Locale | URL |
|---|---|
| en | `/research/index` |
| es | `/es/research/index` |

English → Spanish, starting from `/research/index`:

```
['', 'research', 'index']  →  segments[1] = 'es'  →  ['', 'es', 'index']  →  /es/index
```

`research` is overwritten, and the resulting path does not exist → 404.

Spanish → English, starting from `/es/research/index`:

```
['', 'es', 'research', 'index']  →  segments[1] = 'en'  →  /en/research/index
```

`/en/...` is not a valid URL under `as-needed` either, but next-intl redirects it to `/research/index`, which is why the failure is invisible in that direction.

Worth noting: `pathname` here comes from `next/navigation` (footer.tsx:9), which includes the locale prefix. `usePathname` exported from `@/i18n/config` strips it, and is the locale-aware equivalent.

The same logic is duplicated in `apps/site/src/components/common/locale-switcher/locale-switcher.tsx:22-30`. That component is exported from `src/components/common/index.ts` but is not rendered anywhere today, so it is latent rather than user-visible.

### Steps to Reproduce

1. Open `https://toddagriscience.com/research/index` (English — note there is no locale prefix)
2. Scroll to the footer and open the language menu (bottom right, reads "English")
3. Choose "Español"
4. The browser lands on `/es/index` and renders the 404 page
[...]

PR #1058, opened 2026-08-04
fix(site): localize article dates and correct two spanish strings

### LINKED ISSUE #1056: [Bug]: Spanish locale renders English strings — footer heading and article dates
### Bug Description

On `/es/research/index` (and other Spanish marketing pages), several strings render in English even though the rest of the page is correctly translated. The i18n plumbing itself works — the heading shows "Investigación", the tabs show "Todos" / "Publicación", the footer links show "Investigación", "Acerca de", "Carreras". The failures are specific and have three distinct causes.

**1. Footer column heading "Terms & Policies" — the Spanish message file holds the English string**

`apps/site/src/messages/footer/es.json:9` is a verbatim copy of the English value:

```json
"sections": {
  "todd": "Todd",
  "connect": "Conéctate",
  "legal": "Legal",
  "termsAndPolicies": "Terms & Policies"
}
```

It is rendered at `apps/site/src/app/(unauthenticated)/[locale]/(marketing)/components/footer/footer.tsx:61` via `t('sections.termsAndPolicies')`, so the heading reads "Terms & Policies" in Spanish while every link underneath it ("Términos de Uso", "Política de Privacidad", "Accesibilidad", "Legal") is translated. The sibling heading `sections.todd` is the brand name and `sections.legal` is a valid Spanish word, so those two are fine as-is — only `termsAndPolicies` is wrong.

Note `apps/site/src/app/(unauthenticated)/[locale]/(marketing)/components/footer/footer.test.tsx:38` asserts on the literal `'Terms & Policies'`, so it will need updating alongside the message file.

**2. Article dates never localize — the display locale is hardcoded to `en-GB`**

The row date renders as "1 June 2026" instead of "1 de junio de 2026". The cause is `apps/site/src/lib/sanity/article-display-dates.ts:7`:

```ts
export const ARTICLE_DISPLAY_DATE_LOCALE = 'en-GB';
```

`formatArticleListDate` passes that constant to `toLocaleDateString`, so the route's `[locale]` segment is ignored entirely. The JSDoc on the constant states this is deliberate ("independent of the site `[locale]` route"), which makes this a decision to revisit rather than an oversight — but the result on a Spanish page is an English date.

This is not limited to the research index. The same helper feeds every article surface:

- `components/article-index/article-index.tsx:291` — `/research/index`, `/research/index/[topic]`, `/news`
- `components/news-highlights/news-highlights.tsx:9`
- `src/components/common/news/latest-news-table.tsx:5`
- `src/components/common/news/featured-news-carousel.tsx:6`
- `index/[slug]/utils.ts:17` — article detail hero dates

So one change to the helper fixes the date on all of them at once.

**3. Article title "Introducing Iris" stays in English — the Sanity content model has no localized fields**

`apps/sanity/schemaTypes/article-shared-fields.ts:35` defines `title` as a plain string. There is no `internationalizedArray`, no language field, and no locale-aware querying anywhere in `apps/sanity/schemaTypes/`. Articles are single-language content by design, so `article.title` (rendered at `article-index.tsx:298`) and `article.summary` will always render whatever the editor typed, regardless of the route locale.

This one is a content-model decision, not a code fix, and is considerably larger than the other two. Flagging it here for visibility; it likely belongs in its own issue if the team wants Spanish article content.

### Steps to Reproduce

1. Open `https://toddagriscience.com/es/research/index`
2. Look at the article row — the date reads "1 June 2026" and the title reads "Introducing Iris", while the category above it correctly reads "Publicación"
3. Scroll to the footer — the right-hand column heading reads "Terms & Policies" while all four links beneath it are in Spanish
4. Switch to `https://toddagriscience.com/research/index` (English) — all of the above render identically, confirming the Spanish page is falling back to English strings rather than rendering translated ones

### Expected Behavior

On a `/es/` route, all UI chrome should be Spanish:

- Footer heading: **"Términos y Políticas"** (or whatever wording legal/marketing prefers) instead of "Terms & Policies"
- Article date: **"1 de junio de 2026"** instead of "1 June 2026"

For the date, the fix is to derive the locale from the route rather than hardcoding it — either by passing the active locale into `formatArticleListDate`, or by using `next-intl`'s `useFormatter` / `getFormatter`, which already knows the request locale. The choice matters because the helper is called from both server components and a client carousel, so whichever approach is taken should work in both.

Article titles and summaries are CMS content and are out of scope for a string fix — see item 3.

### Screenshots/Videos

`/es/research/index` in Spanish, showing "1 June 2026" in the article row and "Terms & Policies" in the footer while the surrounding text is correctly translated.

### Browser

Edge

### Device Type

Desktop

### Language/Locale

Spanish (es)

### Additional Context

**The same class of problem is much larger on the legal pages.** Comparing every `es.json` against its `en.json` across all 17 message directories, 214 of 578 keys hold a value byte-identical to the English one. Excluding brand names and acronyms that are legitimately identical (`Todd`, `Legal`, `Iris`, `ESG`), the real gaps are:

| Namespace | Untranslated keys | Notes |
|---|---|---|
| `privacy` | 176 / 178 | The entire privacy policy is in English on `/es/privacy` |
| `terms` | 30 / 30 | The entire terms of use is in English on `/es/terms` |
| `footer` | 1 | `termsAndPolicies` — this issue |
| `investors` | 3 | `ESG` acronym only — not a defect |
| `header` | 1 | `Iris` product name — not a defect |

`/es/privacy` and `/es/terms` therefore serve full English legal documents to Spanish users. That is a translation-content task with legal review attached, not a code change, so it is not part of this issue — but it is the bulk of the untranslated surface and probably warrants its own tracking issue.

Reproduction of the audit:

```bash
cd apps/site/src/messages
# flatten each en.json/es.json pair and report keys whose values are identical
```

**Scope summary for whoever picks this up:** items 1 and 2 are small and self-contained — one JSON value plus its test, and one locale-plumbing change in a shared helper. Item 3 and the legal pages are separate, larger pieces of work.

Orlando275:
## Description

Fixes #1056

Covers items 1 and 2 of the issue. Item 3 (Sanity article titles) and the untranslated legal
pages are content-model and translation-content work, and stay out of scope as the issue sets out.

### Article dates never localized

`ARTICLE_DISPLAY_DATE_LOCALE` was a hardcoded `'en-GB'`, so `/es` routes rendered `1 June 2026`.
`article-display-dates.ts` now maps the route locale to a display locale:

| Site locale | Display locale | Renders |
|---|---|---|
| `en` | `en-GB` | `1 June 2026` (unchanged) |
| `es` | `es-ES` | `1 de junio de 2026` |

English deliberately stays on `en-GB` so no English surface changes.

The locale is an **optional** second argument (`formatArticleListDate(date, locale?)`), falling
back to English, so no call site was forced to change and nothing regresses silently.

How each surface gets its locale:

- `news-highlights`, `latest-news-table`, `featured-news-carousel` are client components → `useLocale()`
- `ArticleIndex` and `CmsArticlePage` are server components that deliberately take their translator
  as a prop rather than reading next-intl context. They take `locale` the same way — an optional
  prop, passed by the four pages that already destructure it from `params`.

### Hero date comma

`formatArticleHeroDate` appends a comma before the year, turning `20 November 2025` into
`20 November, 2025`. Applied to Spanish that produced `20 de noviembre de, 2025`, so the
replacement is now scoped to English output.

### Untranslated strings

- `messages/footer/es.json` — `termsAndPolicies` was a verbatim copy of the English
  `"Terms & Policies"` → `"Términos y Políticas"`
- `messages/header/es.json` — `logIn` read `"Inicio de sesión de cliente"` where the English is
  just `"Login"` → `"Inicio de sesión"`

## Testing

Two new unit test files cover the pure formatters directly: English, Spanish, the fallback when no
locale is supplied, and empty dates.

They test the functions rather than the components on purpose. `vitest.setup.intl.js` mocks the
whole `next-intl` module for every test in the app, with `useLocale` pinned to `'en'` and a
passthrough `NextIntlClientProvider`, so a rendered component cannot observe Spanish. Making that
mock locale-aware would be a test-infrastructure change across all 121 test files and belongs in
its own PR.

Test dates use midday UTC so assertions do not shift with the runner's timezone.

## Notes for review

- No existing test needed changing, and no signature became incompatible.
- The issue predicted `footer.test.tsx` would need updating alongside `footer/es.json`. It does
  not: the global mock always loads the English messages, and `footer/en.json` still reads
  `"Terms & Policies"`.
- Pre-existing and untouched: Sanity stores dates as `"2026-04-20"` with no time, which
  `new Date()` reads as midnight UTC. Rendered in a negative-offset timezone the date shifts back
  a day. Fixing it means passing `timeZone: 'UTC'`, which changes what is on screen, so it is not
  part of this PR.
- Separately filed while working on this: #1057, the language switcher 404s going from English to
  Spanish. It is a routing defect, and it touches `footer.tsx`, which PR #1053 also modifies.


vincenttodd (approved):
except if anybody else has a edit suggestion, looks good to me!

rcjasub (approved):
Review feedback: LGTM 

Scope verified — git diff origin/main...HEAD (had to fetch first, local main was stale) shows exactly the 15 files the PR describes: footer/header string fixes, date-locale helper, and the 6 call sites threading locale through.

Checks run, all clean:
- type-check — 0 errors
- lint — 0 errors (2 pre-existing unrelated  warnings)
- test — 602/602 passing, including 8 new tests for the date helpers

Correctness spot-checks:
- locale param is optional everywhere it was added → no existing call site broke, safe fallback to English for unknown locales
- Hero-date comma fix correctly scoped to en-* only (Spanish phrasing untouched)
- footer.tsx/footer.test.tsx — zero diff vs origin/main, merge from main applied cleanly, no leftover conflict artifacts
- footer.test.tsx still asserts 'Terms & Policies' correctly, since the global test mock always loads en.json

No security or regression concerns — locale comes from the typed route segment, and the lookup defensively falls back rather than trusting it blindly.

PR #1059, opened 2026-08-04
fix(site): keep the page path when switching locale

### LINKED ISSUE #1057: [Bug]: Switching from English to Spanish 404s — the locale prefix overwrites the first path segment

Orlando275:
## Description

Fixes #1057

The footer language menu built the target URL with `segments[1] = newLocale`, which **replaces** the
first path segment instead of adding a locale prefix. That is only correct when a prefix is already
there.

Routing runs with `localePrefix: 'as-needed'` ([i18n/config.ts](apps/site/src/i18n/config.ts)), so
English is served unprefixed and Spanish with `/es`:

```
/research/index  +  Español
  ['', 'research', 'index']  →  segments[1] = 'es'  →  ['', 'es', 'index']  →  /es/index  → 404
```

`research` was overwritten. Every nested marketing page was affected. The homepage was not, because
`/` has no segment to overwrite. The Spanish → English direction only *looked* fine because
next-intl redirected `/en/...` back to the unprefixed path.

### The fix

A shared helper, `localizePathname`, in a new `src/i18n/locale-path.ts` — placed next to `config.ts`
because it implements the `localePrefix: 'as-needed'` rule declared there. It strips a leading
locale segment when present, then re-adds the prefix only for non-default locales.

| From | Choose | Before | After |
|---|---|---|---|
| `/research/index` | Español | `/es/index` → 404 | `/es/research/index` |
| `/es/research/index` | English | `/en/research/index` → redirect | `/research/index` |
| `/` | Español | `/es` | `/es` (unchanged) |
| `/es` | English | `/` | `/` (unchanged) |

The buggy logic existed in **three** places, not two as the issue estimated — twice in `footer.tsx`
(once in `handleLocaleChange`, again in the `useEffect`) and once in the unused `locale-switcher`
component. All three now call the helper so they cannot drift apart again.

### Deliberately left alone

The footer writes the `NEXT_LOCALE` cookie and then does a full `window.location.assign` reload.
That is not an oversight: [proxy/i18n.ts](apps/site/src/proxy/i18n.ts) reads `NEXT_LOCALE` to decide
where to send unprefixed paths, so the reload is what makes the middleware run. Only the path
construction was wrong, so only the path construction changed.

`proxy/special-redirect.ts` also indexes `segments[1]`, but it only reads it for a comparison and is
not affected.

### Tests

- `locale-path.test.ts` — 14 cases covering both directions, the homepage, paths that resemble a
  locale prefix (`/fr/about`, `/es/es/about`), and a round trip.
- `footer.test.tsx` — 4 cases asserting the URL actually handed to `window.location.assign`, plus
  the no-op when the active locale is re-selected.

Both suites were confirmed to fail against the original logic before being committed. Full site
suite: 607 tests passing.

## Checklist


akeight (approved):
LGTM! Thank you! Go ahead and merge.

PR #1060, opened 2026-08-05
fix(site): hide the browser native password reveal control

### LINKED ISSUE #1051: [Bug]: Duplicate password reveal icon on login in Edge
### Bug Description

On `/login`, the password field renders **two eye icons** side by side in Microsoft Edge: the app's own show/hide toggle, plus Edge's native password reveal control.

The app draws its own toggle button, absolutely positioned over the right edge of the input, in `apps/site/src/app/(unauthenticated)/[locale]/(contact)/login/page.tsx:114-138`:

```tsx
<div className="relative">
  <Input
    id="password"
    type={showPassword ? 'text' : 'password'}
    className="border-1 border-[#848484]/80 pr-10 focus:ring-0!"
    ...
  />
  <button
    type="button"
    onClick={() => setShowPassword(!showPassword)}
    className="absolute right-3 top-1/2 -translate-y-1/2 ..."
    aria-label={showPassword ? 'Hide password' : 'Show password'}
    tabIndex={-1}
  >
    {showPassword ? <BiSolidHide className="size-5" /> : <BiShow className="size-5" />}
  </button>
</div>
```

On top of that, Edge (Chromium) injects its own reveal control into every `<input type="password">` through the `::-ms-reveal` pseudo-element. Nothing in the codebase suppresses it — searching the whole repo for `::-ms-reveal` and `::-ms-clear` across `.css`, `.ts` and `.tsx` returns **zero matches** — so both controls paint at the same time and collide inside the `pr-10` padding reserved for the custom button.

The two controls are also independent of each other: Edge's native reveal knows nothing about the React `showPassword` state, so using one leaves the other's icon showing the wrong state.

Chrome, Firefox and Safari don't ship this pseudo-element, which is why only Edge is affected.

### Steps to Reproduce

1. Open `https://toddagriscience.com/login` in Microsoft Edge
2. Look at the right-hand side of the Password field
3. Two eye icons are visible instead of one
4. Click each of them in turn — they toggle the field independently and the icons end up disagreeing about whether the password is visible

### Expected Behavior

Only the app's own show/hide toggle should be visible, and it should be the single source of truth for whether the password is masked.

The native Edge control should be suppressed globally, e.g. in `apps/site/src/app/globals.css`:

```css
input::-ms-reveal,
input::-ms-clear {
  display: none;
}
```

### Screenshots/Videos

_No response_

### Browser

Edge

### Device Type

Desktop

### Language/Locale

None

### Additional Context

**This is not limited to `/login`.** Every masked input in the app pairs a custom toggle with an unsuppressed native control, so all of these show the duplicate icon in Edge:

| File | Line | Field |
|---|---|---|
| `apps/site/src/app/(unauthenticated)/[locale]/(contact)/login/page.tsx` | 118 | Password |
| `apps/site/src/app/(unauthenticated)/signup/components/signup-form.tsx` | 176 | Password |
| `apps/site/src/app/(unauthenticated)/signup/components/signup-form.tsx` | 211 | Confirm password |
| `apps/site/src/app/(authenticated)/(chromeless)/account/reset-password/page.tsx` | 101 | New password |
| `apps/site/src/app/(authenticated)/(chromeless)/account/reset-password/page.tsx` | 133 | Confirm password |

A single global rule in `globals.css` fixes all five at once — no per-component changes needed.

Worth noting while someone is in there: the custom toggle buttons carry `tabIndex={-1}`, which takes them out of the keyboard tab order. Keyboard-only users currently have no way to reach the show/hide control. That may be deliberate, but if it isn't, it's an accessibility gap that belongs in the same fix.

Orlando275:
## Description

Fixes #1051

Edge (Chromium) injects its own reveal control into every `<input type="password">` via the `::-ms-reveal` pseudo-element. Nothing in the repo suppressed it, so every masked input that also ships the app's own show/hide toggle painted two eye icons at once — and because Edge's control knows nothing about the React `showPassword` state, using one left the other's icon showing the wrong state.

### What changed

A `hide-native-password-reveal` class in `globals.css`, applied to the five inputs that already have the app's own toggle:

| File | Field |
|---|---|
| `(unauthenticated)/[locale]/(contact)/login/page.tsx` | Password |
| `(unauthenticated)/signup/components/signup-form.tsx` | Password |
| `(unauthenticated)/signup/components/signup-form.tsx` | Confirm password |
| `(authenticated)/(chromeless)/account/reset-password/page.tsx` | New password |
| `(authenticated)/(chromeless)/account/reset-password/page.tsx` | Confirm password |

`globals.css` is imported by all four route-group layouts, so one rule covers every affected page.

### Deviation from the issue: scoped rather than global

The issue proposes a bare `input::-ms-reveal { display: none }`. That would also hit `(unauthenticated)/accept/components/accept-form.tsx:212,230`, whose two password fields have **no** custom toggle — Edge's native control is their only way to unmask. A global rule would leave that form with no reveal affordance at all, trading one bug for another.

Scoping the rule to a class keeps the accept form working and makes the intent explicit; the comment in `globals.css` records why. If you'd rather have the global rule plus a custom toggle added to the accept form, that's a small follow-up — happy to do it.

(`apps/internal/src/app/login/page.tsx:98` also has an untoggled password field, but it's a separate app with its own `globals.css` and is unaffected either way.)

### Accessibility note from the issue

The login toggle carried `tabIndex={-1}`, which took it out of the tab order — keyboard-only users had no way to reach the show/hide control. The equivalent buttons on signup and reset-password never had it, so it reads as an oversight rather than a deliberate choice. Removed, and covered by a test that fails if it comes back.

### Testing

- `bun validate` passes end to end (606 tests).
- New tests: the class is asserted on the login and reset-password inputs, and the login toggle is asserted to be tab-reachable.
- The `::-ms-reveal` suppression itself can't be asserted in jsdom — Edge is the only engine that ships the pseudo-element, so **the visual check needs a real Edge browser** on `/login`, `/signup` and `/account/reset-password`. One eye icon expected, not two.

## Checklist


rcjasub (approved):
- CSS-only fix, zero runtime cost, scoped correctly (class-based, not global) so the accept-invitation form's untouched password fields still keep their only reveal control.
- Verified independently: globals.css reaches all 4 route-group layouts, the class hits exactly the 5 intended inputs, and the tabIndex={-1} removal on login is safe (toggle button is the next focusable node after the input).
- Ran the affected tests: 15/15 pass. ::-ms-reveal suppression itself can't be tested in jsdom, as the PR notes — still needs a manual check in real Edge before merge.

Non-blocking: signup's confirmPassword field also gets hide-native-password-reveal, but that field has never had its own visible toggle icon (it only inherits showPassword from its sibling). Pre-existing, not introduced here — worth confirming it still has some way to unmask in Edge, but not worth holding up the merge.

PR #1061, opened 2026-08-11
fix(db): verify tls certificate chain on staging and prod connections

### LINKED ISSUE #932: TLS verification disabled on prod and staging DB
## Issue

`rejectUnauthorized: false` is set in both the prod and staging Drizzle configs, opening a MITM path to the database.

## Fix

Set `rejectUnauthorized: true` for prod and staging. Scope `NODE_TLS_REJECT_UNAUTHORIZED=0` to local development only.

Orlando275:
Closes #932

## What the issue asked

1. **Set `rejectUnauthorized: true` for prod and staging.** Both drizzle-kit configs passed `false`, which skips certificate chain validation outright.
2. **Scope `NODE_TLS_REJECT_UNAUTHORIZED=0` to local development only.**

Both are done. Item 2 turned out to be wider than the two config files, see below.

## Changes

A new `packages/db/src/utils/db-ssl.ts` holds the TLS posture for every Postgres connection the package opens, so the app pool, the drizzle-kit configs and the importer scripts cannot drift apart again:

| Helper | Purpose |
| --- | --- |
| `normalizePemCert` | Moved here from `db-target.ts`, where it was a private copy |
| `isLocalDatabaseUrl` | Decides locality from the connection string's hostname |
| `remoteDbSslConfig` | The `verify-ca` posture: validate the chain, skip hostname check |

Applied to:

- `drizzle-prod.config.ts`, `drizzle-staging.config.ts` — the `rejectUnauthorized: false` the issue reports
- `src/schema/connection.ts` — the app pool
- `scripts/create-admin-farm-user.ts` — same hole, not mentioned in the issue
- `scripts/lib/db-target.ts` — was already correct; now shares the helper instead of duplicating it

### On `NODE_TLS_REJECT_UNAUTHORIZED`

This is a **process-wide Node switch**, not a Postgres setting. With it set, certificate verification is off for every outbound TLS call in the process — Supabase, Stripe, OpenAI and Sanity included, not just the database.

All three `env.example` files instructed developers to set it, and four script docblocks used it as a command prefix. Since locality is now derived from the connection string, the variable is not needed anywhere, so it has been removed from the templates and the usage docs rather than merely scoped.

### Why hostname verification stays off

`remoteDbSslConfig` sets `rejectUnauthorized: true` **and** `checkServerIdentity: () => undefined`. That is the `verify-ca` posture the database workflows already use for `pg_dump` (`sslmode=verify-ca`), and the approach `db-target.ts` had already adopted: the managed cluster serves certificates whose subject does not match the host we dial, so full `verify-full` would fail. The chain is still validated against our CA; only the hostname check is waived.

## Verification

`bun validate` passes end to end — 616 tests (592 site, 24 db), including 16 new unit tests for the helpers.

The mechanism was also checked empirically against a local TLS server holding a self-signed certificate whose CN deliberately does not match the dialed host:

| Options passed | `NODE_TLS_REJECT_UNAUTHORIZED` | Result |
| --- | --- | --- |
| none | unset | rejected (`DEPTH_ZERO_SELF_SIGNED_CERT`) |
| none | `0` | **connected, `authorized=false`** |
| `rejectUnauthorized: true` | `0` | rejected |
| previous code (`ca` + `false`) | `0` | **connected, `authorized=false`** |
| this PR (`ca` + `true` + identity waiver) | `0` | connected, **`authorized=true`** |

Rows 3 and 5 are the ones that matter: an explicit `rejectUnauthorized: true` overrides the environment variable, and the posture in this PR still validates the chain while tolerating the hostname mismatch.

Resolved configs were also printed at runtime with the variable set, confirming `rejectUnauthorized=true` for both drizzle configs and `ssl: false` only for `localhost` / `127.0.0.1` / `::1`.

## Not verified

- **No live connection to staging or prod was attempted.** The change follows the posture `db-target.ts` already used in the CI migrator, but a reviewer with cluster access should confirm a real handshake.
- **Whether drizzle-kit forwards `checkServerIdentity` to `pg`.** Its `ssl` field is a zod passthrough object, so unknown keys should survive, but I could not prove it without a TLS-enabled Postgres. Worth noting that **neither `drizzle-prod.config.ts` nor `drizzle-staging.config.ts` is referenced anywhere in the repo** — no package script, workflow or doc invokes them; the CI migrator uses `new Pool(resolveRemoteDbConfig(env))` directly. So if the key were dropped, the blast radius is a manual `drizzle-kit studio` session, not CI, and the outcome would be a loud connection error rather than a silent downgrade.

## Edge cases

`isLocalDatabaseUrl` fails closed: absent, unparseable or host-less connection strings are treated as remote and get full verification. A password carrying a raw `#` or `/` defeats `URL` parsing and therefore surfaces as a TLS error instead of a silent downgrade — percent-encode it, as `create-admin-farm-user` already does. Both cases are covered by tests.


Jose-Gael-Cruz-Lopez:
Closing: superseded by #1092, which merged. `packages/db/src/utils/db-ssl.ts` is on `main` and the prod/staging certificate chain is now verified. #1092 also closed a gap this branch left open — its `declaredPlaintextHosts()` did not subtract the managed clusters, so `DATABASE_PLAINTEXT_HOSTS=<prod host>` still yielded `ssl:false`. Thanks for the original work on #932, it drove the fix that landed.

PR #1063, opened 2026-08-11
Revert "Fix 1032 contact redesign"

vincenttodd:
Reverts toddagriscience/nightcrawler#1054


Jose-Gael-Cruz-Lopez:
## This reverts the security-audit fix that is already on main, which is why Security Audit is red

The `Security Audit` job (`.github/workflows/security.yml:26`) runs exactly one assertion: `bun audit --audit-level=high`. This PR breaks it in two mutually reinforcing ways, both visible in the diff.

### 1. It deletes the audit suppressions

```diff
-        run: bun audit --audit-level=high --ignore=GHSA-w3rx-r6r6-pgpr --ignore=GHSA-5p2g-fcmc-qvqq
+        run: bun audit --audit-level=high
```

### 2. It simultaneously rolls back the `overrides` that remediate the underlying advisories

Root `package.json` — the block whose whole purpose is forcing patched transitive versions:

```diff
-    "lodash": "^4.18.1",      +    "lodash": "^4.17.24",
-    "lodash-es": "^4.18.1",   +    "lodash-es": "^4.17.24",
-    "js-yaml": "^4.3.1",      +    "js-yaml": "^4.3.0",
-    "nanoid": "^5.1.16",           (deleted outright)
-    "esbuild": "^0.28.2",     +    "esbuild": "^0.28.1",
-    "minimatch": "^10.2.6",   +    "minimatch": "^10.2.3",
-    "rollup": "^4.62.4",      +    "rollup": "^4.59.0",
```

Plus `storybook ^10.5.7 → ^10.4.4` in both `package.json` and `bun.lock`. That is **verbatim the change `bca54ad` (#1064, "fix: resolve storybook security audit failures") landed on `main` to fix this exact check.** The ignore flags are removed at the same moment the advisories they were suppressing come back.

### 3. `bun.lock` ends up disagreeing with `package.json`

| Package | `package.json` | `bun.lock` |
| --- | --- | --- |
| `undici` | `^7.29.0` | `^7.27.2` |
| `eslint-config-next` | `16.2.12` | `16.2.9` |

An independent way this breaks CI, separate from the audit.

### 4. The scope is much wider than "contact redesign"

Because it was generated as a whole-commit revert, it also drags back:

- `apps/site/src/proxy/auth.ts` — **auth middleware**, reverting route-protection logic
- `packages/db/src/utils/get-embedding.ts` — back to `apiKey ?? 'NOTAKEY'`, which converts a named configuration error into a silent 401 at call time (`main`'s lazy `embeddingsClient()` is strictly better; see #1007)
- `packages/db/src/utils/preview-form-submission-hydration.ts`, `packages/db/scripts/seed-local-db.ts`, `apps/site/src/i18n/message-files.ts`, the footer
- Pure Prettier-only reflows in `apply/types.ts`, `account/db.ts`, `reminders/types.ts` — zero behavioural content, pure conflict surface

### 5. It also contends with #960

It restores `(contact)/contact/actions.ts`, which #960 patches and which `main` deleted. Whichever lands first forces the other to be rewritten.

### Recommendation

Close. If specific behaviour from the old contact page is genuinely wanted back, a narrow PR touching only the contact route — re-implemented against the current `/forms/[slug]` Sanity architecture, with no `package.json`, `bun.lock`, workflow or auth-middleware changes — would be far safer than a 26-file blanket revert.

Note `main`'s contact experience is now the Sanity `DynamicForm` flow; `(contact)/contact/` holds only `layout/page` files and `submitPublicInquiry` has zero occurrences repo-wide. Resurrecting a Google-Sheets-backed contact action alongside the DB-backed form flow would reintroduce a second, divergent submission path.

PR #1064, opened 2026-08-11
fix: resolve storybook security audit failures

yjm7vj:
## Description

Summary
Replaced Storybook’s Next/Vite adapter with the React/Vite adapter to remove the vulnerable dependency path causing the security audit failure.
Removed the previous audit ignore workaround so bun audit --audit-level=high runs normally.
Added a small Storybook-only mock for next/font/local so Storybook still builds with the new adapter.

Validation
bun audit --audit-level=high
bun run --filter @nightcrawler/site type-check
bun run --filter @nightcrawler/site build-storybook

## Checklist

### DIFF
--- a/.github/workflows/security.yml
+++ b/.github/workflows/security.yml
@@ -24,4 +24,4 @@ jobs:
         run: bun i
 
       - name: Check for security vulnerabilities
-        run: bun audit --audit-level=high --ignore=GHSA-w3rx-r6r6-pgpr --ignore=GHSA-5p2g-fcmc-qvqq
+        run: bun audit --audit-level=high
--- a/.prettierignore
+++ b/.prettierignore
@@ -3,7 +3,6 @@
 **/out
 **/build
 **/next-env.d.ts
-**/.storybook
 **/storybook-static
 **/package-lock.json
 **/.sanity
--- a/apps/site/.storybook/decorators/withFonts.tsx
+++ b/apps/site/.storybook/decorators/withFonts.tsx
@@ -1,3 +1,5 @@
+// Copyright © Todd Agriscience, Inc. All rights reserved.
+
 import React from 'react';
 import { neueHaasUnica, utahWGLCondensed } from '../../src/lib/fonts';
 
--- a/apps/site/.storybook/decorators/withNextNavigation.tsx
+++ b/apps/site/.storybook/decorators/withNextNavigation.tsx
@@ -1,0 +1,34 @@
+// Copyright © Todd Agriscience, Inc. All rights reserved.
+
+import React from 'react';
+import { __setNavigation } from '../mocks/next-navigation';
+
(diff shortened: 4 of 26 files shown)


rcjasub (commented):
Just Realized @Jose-Gael-Cruz-Lopez  address this above 

Blocking: account-side-menu.stories.tsx crashes in Storybook. AccountSideMenu calls useRouter()/usePathname() from next/navigation, which @storybook/nextjs-vite used to auto-mock; @storybook/react-vite doesn't, and this PR only adds an alias for next/font/local. Reproduced locally: rendering the component without a next/navigation mock throws TypeError: Cannot read properties of null (reading 'replace') at usePathname().replace(...). Needs a next/navigation mock added before this merges.

Separately, non-blocking: preview.ts/vitest.setup.ts still import from @storybook/nextjs, which was never a dependency — should be @storybook/react-vite. Currently inert (nothing executes those files), so it's fine as a fast-follow.

Jose-Gael-Cruz-Lopez:
@rcjasub Both addressed — #1066 is merged into this branch, so the fixes are on it now.

**Blocking (`next/navigation`):** fixed. `.storybook/mocks/next-navigation.ts` is aliased in `viteFinal` and fed by a `withNextNavigation` decorator that reads `parameters.nextjs.navigation`, so `account-side-menu.stories.tsx` keeps its existing `pathname: '/account'` unchanged. Verified in a running Storybook: the menu renders and "Farm information" is correctly marked active.

**Non-blocking (`@storybook/nextjs`):** fixed too — `preview.ts` and `vitest.setup.ts` now import from `@storybook/react-vite`. There was a third instance you didn't hit: `decorators/withStorybookProvider.tsx` imported `@storybook/react`, equally undeclared. One small correction on "currently inert" — Storybook loads `preview.ts` on every start, so on a clean install with nothing hoisting that package it would have failed to boot rather than sit dormant.

While reproducing yours, three more of the same class turned up:

- `next/image` fell back to Next's optimizer, pointing every image at `/_next/image`, which Storybook doesn't serve.
- `next/link` reads `process` at module scope, so **any** story containing a link died with `ReferenceError: process is not defined` — including via next-intl's `Link`. That's what was blanking UI/Header.
- The font mock returned the CSS custom-property name where the real loader returns a generated class, so `--font-neue-haas` was never defined and every story silently rendered in Arial. That one would have landed as a whole-suite Chromatic diff.

Underlying reason none of this could fail CI: `.storybook` sits outside type-check, lint, and format. It's now type-checked and format-checked, which immediately surfaced two more real errors (a `readonly` alias mismatch and an unsupported `showName` toolbar property).

Still worth someone opening the preview Storybook before approving — I checked AccountSideMenu, Header, NewsCard, and the font variables by hand, but there are more stories than that.

vincenttodd (approved):

PR #1067, opened 2026-08-18
fix(site): restore careers disclosures with accommodation inquiry link

### LINKED ISSUE #590: Add an inquiry link to /career disclosure footer for disability accommodation

Kenyang1:
## Description

Fixes #590

The careers legal disclosures exposed a raw support mailbox for disability-accommodation requests. While fixing that, I found the situation was a bit worse than the issue describes: the only component that rendered those disclosures (`careers-externship.tsx`) has been **unimported since the careers refactor in #886**, so the live `/careers` page has been shipping **no legal disclosures at all** — no EEO statement, no accommodation notice, and no Transparency in Coverage link. (#1011 deletes that dead file, so I left it untouched to avoid a conflict.)

This PR restores the disclosure block on the live `CareersLandingView` and turns every contact point into a proper link instead of scrapeable text:

- **`Disclaimer`** (shared component) now renders messages through next-intl `t.rich` with a `links` prop (tag → href). Root-relative hrefs render as locale-aware `<Link>`s; absolute `http(s)` hrefs open in a new tab with `rel="noopener noreferrer"`. Props moved to `types/disclaimer.ts`; unit tests and a Storybook story added.
- **`careers.disclaimers` (en + es)**: the email is replaced by an `<inquiry>` tag pointing at `/contact`. Since this PR makes paragraphs 3–4 visible for the first time, the Transparency in Coverage URL and the Privacy Policy reference are also linked (`<coverage>` / `<privacy>`) rather than shipping as a bare URL / plain text.
- **`vitest.setup.intl.js`**: the global next-intl mock had no `t.rich` and could not resolve dotted namespaces such as `careers.disclaimers` (the exact limitation the component's old comment complained about). Both fixed, shared between the `useTranslations` and `getTranslations` mocks.
- **`.husky/commit-msg`** (separate commit, happy to split into its own PR): the hook called `npx --no -- commitlint`, which cannot resolve packages from a Bun-installed `node_modules` tree (every commit was rejected before commitlint ran, observed on Windows). Switched to `bunx commitlint`, matching `pre-commit`'s `bunx lint-staged`. Verified the hook still rejects non-conventional messages.

### Verified

- `tsc`, `eslint`, `prettier` clean; full `apps/site` vitest suite green (123 files / 601 tests).
- Dev server, both locales: `/careers` renders the 5 disclosures with `href="/contact"` (en) / `href="/es/contact"` (es), external link has `target="_blank" rel="noopener noreferrer"`, and there is no email address or bare URL anywhere in the page text.

## Checklist

image

### DIFF
--- a/.husky/commit-msg
+++ b/.husky/commit-msg
@@ -1,4 +1,4 @@
 # Enforce conventional commit format
 # Examples: "feat: add header", "fix: resolve login issue", "docs: update README"
 cd ./apps/site
-bunx commitlint --edit "$1"
+bunx @commitlint/cli --edit "$1"
--- a/.husky/commit-msg
+++ b/.husky/commit-msg
@@ -1,4 +1,4 @@
 # Enforce conventional commit format
 # Examples: "feat: add header", "fix: resolve login issue", "docs: update README"
 cd ./apps/site
-npx --no -- commitlint --edit $1
+bunx commitlint --edit "$1"
--- a/apps/site/src/app/(unauthenticated)/[locale]/(marketing)/careers/components/careers-externship.tsx
+++ b/apps/site/src/app/(unauthenticated)/[locale]/(marketing)/careers/components/careers-externship.tsx
@@ -3,6 +3,10 @@
 'use client';
 
 import { Disclaimer } from '@/components/common/disclaimer/disclaimer';
+import {
+  CAREERS_DISCLAIMER_COUNT,
+  CAREERS_DISCLAIMER_LINKS,
+} from '../constants/careers-disclaimers';
 import HeaderImg from '@/components/common/header-img/header-img';
 import { motion, useScroll, useTransform } from 'framer-motion';
 import { useTranslations } from 'next-intl';
@@ -114,7 +118,8 @@ export function CareersExternship() {
           <div className="text-left">
             <Disclaimer
               translationLoc="careers.disclaimers"
-              disclaimerCount={5}
+              disclaimerCount={CAREERS_DISCLAIMER_COUNT}
+              links={CAREERS_DISCLAIMER_LINKS}
             />
           </div>
         </main>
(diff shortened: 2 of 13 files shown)


.husky/commit-msg
```diff
@@ -1,4 +1,4 @@
 # Enforce conventional commit format
 # Examples: "feat: add header", "fix: resolve login issue", "docs: update README"
 cd ./apps/site
-bunx commitlint --edit "$1"
+bunx @commitlint/cli --edit "$1"
```
akeight, line 4: Was this not working?

apps/site/public/go/creatorsbig.mp4
akeight, line 1: why was this removed in this PR?

apps/site/public/go/creatorssmall.mp4
akeight, line 1: same with this?

.husky/commit-msg
Kenyang1: Good question — the original `npx --no -- commitlint` was genuinely broken on Windows with our Bun-managed `node_modules`: every commit was rejected before commitlint even ran, with

```
npx canceled due to missing packages and no YES option: ["commitlint@21.2.2"]
```

Two things going on there: Bun hoists everything to the repo root and, on Windows, writes `.exe`/`.bunx` shims that npx doesn't recognize as a local install. So npx mapped the bare command name to the standalone `commitlint` registry package — unpinned, and not the `@commitlint/cli@21.2.1` from our lockfile — and `--no` (correctly) refused to download it, failing the hook. It presumably kept working on macOS/Linux because Bun creates plain symlink bins there that npx can execute.

`bunx` fixes the resolution (matching the `bunx lint-staged` we already use in pre-commit), and the follow-up pin to `@commitlint/cli` closes the remaining gap: if `node_modules` were ever missing, `bunx commitlint` could auto-fetch that other standalone `commitlint` package, while `bunx @commitlint/cli` can only ever resolve the right one.

Verified on the current head: malformed message → hook exits 1 and blocks the commit, conventional message → exits 0.

It should be good to revert it back to normal if it doesn't affect the company. It might have been the version of the Windows I had that kept on failing locally. Thats why the change was made

akeight:
Issue was taken care of in a previous merge and not closed.

PR #1068, opened 2026-08-20
fix(site): restore jsdom web storage when node shadows it in tests

naman-bajpai:
## Description

No linked issue — found while running `bun validate` on a local checkout.

On Node 22+ the site test suite fails with 19 errors across 6 files, none of
which CI sees. The repo targets Node 20 (`.nvmrc`, and `testing.yml` pins
`node-version: '20'`), so this only bites contributors whose local Node is
newer.

### Root cause

Node 22 shipped a built-in `localStorage`/`sessionStorage`, and from Node 24 it
is exposed by default. It only functions when the process is started with a
valid `--localstorage-file` path; otherwise Node installs a global with no
methods on it and emits:

```
Warning: `--localstorage-file` was provided without a valid path
```

That global takes precedence over the one jsdom installs, so `localStorage` is
present but inert:

```console
$ node -e "console.log(typeof localStorage, typeof localStorage?.clear)"
object undefined
```

### Why one broken global caused 19 failures

`localStorage.clear()` is almost always called from `beforeEach`/`afterEach`.
The throw aborts the hook chain before Testing Library's `cleanup` runs, so
mounted DOM survives into the next test and it fails with `Found multiple
elements with the role "button"`. Only 5 of the 19 failures were storage
assertions; the rest were collateral in `sidebar.test.tsx` and
`sidebar-collapse-toggle.test.tsx`.

### Fix

Install a working in-memory `Storage` in `vitest.setup.ts` when the ambient one
is unusable. The guard is inert on Node 20 and in CI — it only engages on newer
local toolchains, so contributors see the same results locally that CI does.

The helper uses a plain record and the `delete` operator rather than a `Map`,
because `drizzle/enforce-delete-with-where` is configured repo-wide without a
`drizzleObjectName` and flags any `.delete()` method call, database-related or
not. Narrowing that rule would be a reasonable follow-up.

### Verification

- `bun run validate` passes end to end (format:check → type-check → lint → test → build).
- Site suite goes from **6 files / 19 tests failing** to **124 files / 608 tests passing** on Node 25.8.1.
- No production code touched — the change is confined to test setup.

## Checklist

### DIFF
--- a/apps/site/vitest.setup.ts
+++ b/apps/site/vitest.setup.ts
@@ -2,3 +2,72 @@
 
 import '@testing-library/jest-dom';
 import 'whatwg-fetch';
+
+/**
+ * Restores a working Web Storage implementation when the host Node build
+ * shadows jsdom's.
+ *
+ * Node 22 shipped a built-in `localStorage`/`sessionStorage`. From Node 24 it
+ * is exposed by default, but it only functions when the process is started
+ * with a valid `--localstorage-file` path; otherwise Node installs a global
+ * with no methods on it and warns. That global takes precedence over the one
+ * jsdom installs, so `localStorage.clear()` throws "is not a function" inside
+ * tests. Because those calls usually live in `beforeEach`/`afterEach`, the
+ * throw aborts the hook chain before Testing Library's `cleanup` runs, leaving
+ * mounted DOM behind and turning a single failure into cascading "Found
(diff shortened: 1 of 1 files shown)


akeight:
Fixed with the rebase from `merge-ready/1068-vitest-storage-shim`

PR #1069, opened 2026-08-20
fix(site): stop management zone updates from writing identity columns

### LINKED ISSUE #997: Allow clients to add pending management zones before soil sample kit is processed
### Problem Statement

When a farm has zero management zones, the dashboard (apps/site/src/app/(authenticated)/(shell)/(widgets)/page.tsx:39-76) shows a static empty state: "Your platform is ready when you are... there are no management zones on this account yet," with links to Orders/Contact/Account only. There's no way for the client to request or add a zone from here.

### Context

Client sees a dead-end message and must contact support out-of-band to get a management zone created. Creating a management zone isn't instant — it requires physically sending the client a soil sample kit and processing the results before a zone can be considered active — so this needs to be modeled as a request/pending state rather than an immediate creation flow. This directly follows up on the empty state introduced in the widgets dashboard page.

### Constraints

This issue should not:
- Change the schema or behavior of active/existing management zones
- Automatically create an active zone without soil sample processing
- Introduce breaking changes to the current dashboard empty state for accounts that already have zones

### Technical Details

- Likely touches apps/site/src/app/(authenticated)/(shell)/(widgets)/page.tsx (empty state) and the managementZone schema (packages/db/src/schema)
- Needs either a status column on managementZone (e.g. pending / active) or a separate "zone request" table/entity
- Needs an ops-facing way to view and process pending requests (mark as sample-kit-sent, mark as processed/active)

### User Story

As a client with no management zones yet, I want to submit a request to add a management zone, so I can start the soil sampling process instead of hitting a dead end on the dashboard.

### Use Case

- Client logs in and sees the "no management zones" empty state
- Client clicks a new "Add management zone" action instead of just Orders/Contact/Account
- Client submits basic info for the zone request (e.g. location/acreage)
- Request is created in pending status; client sees it listed as pending on the dashboard
- Ops receives/sees the pending request, ships a soil sample kit, and processes results
- Once processed, the zone flips to active and behaves like a normal management zone on the dashboard

naman-bajpai:
## Description

Related to #997 — this is the independent fix @henryng15 flagged in
https://github.com/toddagriscience/nightcrawler/issues/997#issuecomment-4927622590
("that fix is independent of this issue and could ship as its own PR"). It does
**not** close #997, which is still blocked on the two product decisions in that
thread.

> Stacked on #1068, which this branch needs in order for `bun validate` to pass
> on Node 22+. Base retargets to `main` automatically once that merges. The only
> commit belonging to this PR is `fix(site): stop management zone updates from
> writing identity columns`.

### The problem

`updateManagementZone` spread the client payload straight into `.set()`:

```ts
.set({ ...input })   // input is ManagementZoneInsert — the full row type
```

The edit form seeds react-hook-form's `defaultValues` from the whole zone row,
so every submission posts `id`, `farmId`, `createdAt` and `updatedAt` back to
the server. The `.where()` clause scopes *which* row is updated to the caller's
farm, but places no constraint on *what* is written to it — and both `id` and
`farmId` are writable columns on `management_zone`. A farm Admin could therefore
reassign their own zone to another farm via `farmId`, or collide a primary key
via `id`.

The internal app's equivalent (`apps/internal/.../farms/actions.ts`) is already
typed to an explicit field list, so this is confined to the customer-facing
action. `createManagementZone` and `deleteManagementZone` were already safe.

### The fix

Narrow the payload to the seven columns the form owns. Two details follow from
how the form actually behaves:

- `undefined` is dropped so untouched fields keep their stored value.
- `null` is preserved — that is how the form clears an optional date via
  `valueAsDate`.

Filtering can now empty the payload, so that case is guarded with a clear error
rather than letting drizzle throw its generic "no values to set".

### Verification

The tests fail against the vulnerable code — reverting the action to
`.set({ ...input })` fails 5 of the 12, including both mass-assignment guards.
They mock the drizzle query builder, so no database is required to run them.

`bun validate` passes end to end.

### Noticed but not fixed

`management-zone-form.tsx` registers `location.0` / `location.1` on
`type="number"` inputs without `valueAsNumber`, so coordinates reach the server
as strings against a `point({ mode: 'tuple' })` column. Pre-existing and
unrelated; left alone to keep this diff reviewable, but worth its own issue.

## Checklist

### DIFF
--- a/apps/site/src/app/(authenticated)/(no-sidebar)/account/(with-shell)/management-zones/[zone]/actions.test.ts
+++ b/apps/site/src/app/(authenticated)/(no-sidebar)/account/(with-shell)/management-zones/[zone]/actions.test.ts
@@ -106,6 +106,10 @@ function expectFarmScopedWhere(captured: unknown, zoneId: number) {
 /**
  * Builds a payload shaped like a real form submission, which round-trips the
  * whole stored row back to the server alongside the edited fields.
+ *
+ * `location` is a numeric tuple because that is the shape the column expects,
+ * which is also what the form posts now that both coordinate inputs register
+ * with `valueAsNumber`.
  */
 function makeFormPayload(
   overrides: Partial<ManagementZoneInsert> = {}
@@ -236,7 +240,12 @@ describe('updateManagementZone', () => {
   });
 
   it('keeps the update scoped to the zone and the caller’s farm', async () => {
-    await updateManagementZone(ZONE_ID, makeFormPayload());
+    // The client id diverges from the session id, so a WHERE built from
+    // `input.farmId` instead of the session would bind 99 and fail below.
+    await updateManagementZone(
+      ZONE_ID,
+      makeFormPayload({ farmId: OTHER_FARM_ID })
+    );
 
     expect(mockUpdate).toHaveBeenCalledTimes(1);
     expectFarmScopedWhere(capturedWhere.value, ZONE_ID);
--- a/apps/site/src/app/(authenticated)/(no-sidebar)/account/(with-shell)/management-zones/[zone]/actions.test.ts
+++ b/apps/site/src/app/(authenticated)/(no-sidebar)/account/(with-shell)/management-zones/[zone]/actions.test.ts
@@ -1,0 +1,311 @@
+// Copyright © Todd Agriscience, Inc. All rights reserved.
+
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
+import { and, eq } from 'drizzle-orm';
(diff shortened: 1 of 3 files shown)


Jose-Gael-Cruz-Lopez:
Closing: superseded by #1093, which merged. `actions.ts` on `main` now has `pickEditableManagementZoneFields`, excluding `id`/`farmId`/`createdAt`/`updatedAt`, so the mass-assignment path is closed. #1093 also fixed an adjacent data-loss bug this branch missed — the form seeded `location: [0, 0]`, overwriting real coordinates on every save. That is now `zone.location ?? [0, 0]`.

PR #1072, opened 2026-08-21
fix(db): finish the verify-full escalation and name the right cert var

Jose-Gael-Cruz-Lopez:
## Description

Stacked on #1061 — merge into `fix/932-tls-verification` before that PR lands.

#1061 closes a real hole (`connection.ts:14` disables TLS entirely whenever
`NODE_TLS_REJECT_UNAUTHORIZED` is set, and both drizzle configs ship
`rejectUnauthorized: false`). Two gaps remain.

### 1. The verify-full escalation stops short of CI

`db-reconcile.yml` already dials both clusters at `verify-full` (lines 116, 250,
308) with a comment saying a downgrade needs explicit justification. #1061 drops
the `checkServerIdentity: () => undefined` override in `db-target.ts`, moving the
app pool and the migration scripts to verify-full too. But these two were left at
`verify-ca`:

| File | Line | Was |
| --- | --- | --- |
| `.github/workflows/pr-database-compatibility-check.yml` | 71 | `sslmode=verify-ca` (staging) |
| `.github/workflows/release-database-compatibility-check.yml` | 69 | `sslmode=verify-ca` (prod) |

That matters beyond consistency: `migrate-remote-db.ts:34` is the CI migration
path and hostname verification is **new** for it under #1061. Raising these two
makes CI prove the posture against staging and prod *before* the app pool starts
depending on it. `db-reconcile.yml` passing today is the evidence the provider
certs can satisfy it.

Local dumps stay `sslmode=disable` — the docker cluster serves no TLS.

### 2. A missing staging cert named the wrong variable

`resolveDbPoolConfig` called `remoteDbSslConfig(rawCert)` with no `certVarName`,
so it fell back to the `DATABASE_PEM_CERT` default. But `caCertForConnection`
host-matches staging and resolves through `stagingCaCert()`, which reads
`STAGING_DATABASE_PEM_CERT` first. A staging URL with no cert therefore threw
telling you to set a variable that was not the one consulted — in the one code
path whose whole purpose is arriving named rather than as an opaque
`SELF_SIGNED_CERT_IN_CHAIN`.

Extracted `caCertSourceForConnection`, returning the cert and the variable that
supplied it, so the host match is not duplicated. `caCertForConnection` keeps its
exported signature and all its existing tests.

## Verification

- `packages/db`: **50 tests / 6 files pass**, including `db-ssl.test.ts`
- `tsc --noEmit -p packages/db/tsconfig.json`: clean

## Not addressed here

`DATABASE_PLAINTEXT_HOSTS` still lets any named host drop to `ssl: false`. Much
narrower than the switch it replaces, but consider ignoring it when
`NODE_ENV === "production"`, or rejecting entries equal to
`STAGING_DATABASE_HOST` / `PROD_DATABASE_HOST`.

## Checklist

### DIFF
--- a/.github/workflows/pr-database-compatibility-check.yml
+++ b/.github/workflows/pr-database-compatibility-check.yml
@@ -67,8 +67,12 @@ jobs:
             > /tmp/staging-db-ca.pem
           openssl x509 -in /tmp/staging-db-ca.pem -noout -subject
 
+          # sslmode=verify-full also validates the server certificate's hostname
+          # against STAGING_DATABASE_HOST (verify-ca only validates the CA
+          # chain), matching db-reconcile.yml. Keep this in step with the TLS
+          # posture the app pool and drizzle configs use.
           PGPASSWORD="$STAGING_DATABASE_PASSWORD" /usr/lib/postgresql/18/bin/pg_dump \
-            "host=$STAGING_DATABASE_HOST port=$STAGING_DATABASE_PORT user=$STAGING_DATABASE_USER dbname=$STAGING_DATABASE_DATABASE sslmode=verify-ca sslrootcert=/tmp/staging-db-ca.pem" \
+            "host=$STAGING_DATABASE_HOST port=$STAGING_DATABASE_PORT user=$STAGING_DATABASE_USER dbname=$STAGING_DATABASE_DATABASE sslmode=verify-full sslrootcert=/tmp/staging-db-ca.pem" \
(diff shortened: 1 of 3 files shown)


vincenttodd:

Jose-Gael-Cruz-Lopez:
Closing: superseded by #1092, which merged. The `verify-full` escalation and the certificate variable naming this branch addressed are both on `main`.

PR #1033, opened 2026-07-29
feat(site): add back navigation to account subpages

Orlando275:
## Description

Fixes #763

Adds optional `backHref` and `backLabel` props to `AccountInfo`, the wrapper shared by every `/account` page. Each subpage passes its parent route; the `/account` index omits the prop, since the account header already links home.

Reuses the existing back-arrow pattern from `account-header.tsx` (`BiArrowBack` from react-icons). The icon is marked `aria-hidden` since the adjacent text carries the meaning.

## Checklist

- [x] You've included unit or integration tests for your change, where applicable.
- [x] You've included inline docs for your change, where applicable.
- [x] Any components that you've modified are accessible.
- [x] You've used conventional commits where appropriate

PR #1073, opened 2026-08-21
fix(site): use the react-vite storybook framework and drop the duplicate back link

Jose-Gael-Cruz-Lopez:
## Description

Stacked on #1033 — merge into `feat/763-back-in-account-subpages` before that PR lands.

### 1. Storybook framework import will fail CI

`account-info.stories.tsx:3` imported `@storybook/nextjs-vite`. Main migrated to
`@storybook/react-vite` in `bca54ad`, and `nextjs-vite` now has **no entry in
`apps/site/package.json` and zero occurrences in `bun.lock`**.

`apps/site/tsconfig.json:25-33` includes `**/*.tsx`, so both `tsc --noEmit`
(`code-quality.yml:33`) and `next build` / `build-storybook` (`build.yml:32,35`)
resolve story imports and fail on TS2307. `eslint.config.mjs:49` globally ignores
`*.stories.*`, so `lint` stays green — which makes the failure look stranger than
it is.

> Worth flagging: **#1033 currently reports 5/5 green.** Those checks ran
> `2026-07-29`; the storybook migration landed `2026-08-11`. The green status
> predates the breaking change by 13 days and will flip the moment CI re-runs.

This PR only fixes the one story file #1033 adds. The other stories on this branch
still naming `nextjs-vite` are untouched by #1033, so they resolve from `main` on
merge.

### 2. Duplicate back link on the zone detail page

#1033 adds `backHref` to `management-zones/[zone]/page.tsx`, so `AccountInfo`
renders a Back link at the top. But `management-zone-form.tsx:200-205` already
rendered its own `Back to zones` link to the **same route** at the bottom.

Removed the in-form link so `AccountInfo` owns back navigation everywhere, and
carried its more specific wording onto the page as `backLabel="Back to zones"` —
which also avoids adding one more link whose accessible name is just `Back`. In a
screen-reader link list across the account area those are indistinguishable;
consider destination-specific labels on the other five call sites too.

`Link` and `BiArrowBack` were only used by the removed block, so both imports go
with it.

## Verification

- `tsc --noEmit` in `apps/site`: clean
- `vitest run` over the account subtree: **13 tests / 7 files pass**

## Note

`management-zone-form.tsx` is also rewritten by #1035 (shadcn calendar). Landing
#1033 first keeps that rebase to the import block.

## Checklist

### DIFF
--- a/apps/site/src/app/(authenticated)/(no-sidebar)/account/(with-shell)/management-zones/[zone]/management-zone-form.tsx
+++ b/apps/site/src/app/(authenticated)/(no-sidebar)/account/(with-shell)/management-zones/[zone]/management-zone-form.tsx
@@ -11,9 +11,7 @@ import type {
   ManagementZoneSelect,
 } from '@/lib/types/db';
 import { formatActionResponseErrors } from '@/lib/utils/actions';
-import Link from 'next/link';
 import { useForm } from 'react-hook-form';
-import { BiArrowBack } from 'react-icons/bi';
 import { updateManagementZone } from './actions';
 
 export default function ManagementZoneForm({
@@ -197,12 +195,6 @@ export default function ManagementZoneForm({
               <p className="text-sm text-[#ff4d00]">{errors.root.message}</p>
             </div>
           )}
-          <Link
-            href={'/account/management-zones'}
-            className="flex flex-row items-center gap-2 text-sm font-light hover:text-foreground/70 transition-all duration-300 ease-in-out"
-          >
-            <BiArrowBack className="size-5" /> Back to zones
-          </Link>
         </div>
       </div>
     </form>
--- a/apps/site/src/app/(authenticated)/(no-sidebar)/account/(with-shell)/management-zones/[zone]/page.tsx
+++ b/apps/site/src/app/(authenticated)/(no-sidebar)/account/(with-shell)/management-zones/[zone]/page.tsx
@@ -46,6 +46,7 @@ export default async function ManagementZonePage({
           : 'View management zone details.'
       }
       backHref="/account/management-zones"
+      backLabel="Back to zones"
     >
       <ManagementZoneForm
         zone={curManagementZone}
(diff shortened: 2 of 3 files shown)


Jose-Gael-Cruz-Lopez:
## Why CI is red on this PR, and why the change is still correct

Being upfront: the `Build Application` and `Code Quality Checks` failures here are caused by my own storybook import change, in a way that resolves itself only after the base branch is rebased. Detail so nobody wastes time on it.

The failure is:

```
error TS2307: Cannot find module '@storybook/react-vite'
```

That is the **inverse** of the failure on the base branch. The reason is dependency resolution:

| Context | `package.json` / `bun.lock` has | Correct import |
| --- | --- | --- |
| This base branch today | `@storybook/nextjs-vite` | `nextjs-vite` |
| `main` (since `bca54ad`) | `@storybook/react-vite` | `react-vite` |

This PR merges into the base branch, whose lockfile predates the migration — so `react-vite` genuinely is not installed there yet, and CI fails.

**After the base branch merges to `main` the situation reverses.** The base branch does not modify `package.json`, so `main`'s version wins on merge: `react-vite` is present and `nextjs-vite` is gone. At that point `nextjs-vite` is the broken import and `react-vite` is the correct one.

### So the ordering is

1. Rebase the base branch onto current `main` (it needs this regardless — it is stale, and its other story files that still name `nextjs-vite` resolve from `main` on merge)
2. Merge this PR

After step 1, CI here goes green and the import is correct for the destination.

### If you would rather not carry a red PR

Take only the non-storybook half of this branch and change the import yourself during the rebase. That half is independent and useful on its own. I would rather flag the tradeoff than quietly leave an import that breaks the moment the branch lands on `main`.

### Separately: `Security Audit` is failing everywhere right now

Not related to this PR. `GHSA-r292-9mhp-454m` (node-tar, affects `<=7.5.20`) was published after `main`'s last green run, and `main`'s `overrides` pinned `tar: ^7.5.19` which resolved to exactly `7.5.20`. I opened #1082 against `main` to bump it — `bun audit --audit-level=high` exits 0 with that change.

Stale branches will still fail Security Audit after #1082 for a *different* reason: branches cut before `bca54ad` (#1064) predate the `overrides` block remediating the nanoid/undici advisories, so their merge refs resolve an old tree. That one needs a rebase, not a dependency change.

vincenttodd:

akeight:
We don't need storybook, so if that's the blocker just remove it from this work

Jose-Gael-Cruz-Lopez:
Closing: superseded by #1094, which merged. The react-vite Storybook framework fix and the duplicate back-link removal are both on `main`.

PR #1075, opened 2026-08-21
fix(ci): pin sync-imps-to-staging actions to immutable commit SHAs

Jose-Gael-Cruz-Lopez:
## Description

Targets `main` directly — independent of #1022.

`sync-imps-to-staging.yml` is the **only workflow still using mutable action tags**, and it is the most privileged one in the repo.

### Why it was missed

It is not an oversight in #1022. The file was added to `main` by `2f2060a` (#1026) **after** that PR's branch was cut, so it was never in scope. When #1022 merges, this file arrives unpinned regardless — hence a separate PR.

### Why it matters most

| | |
| --- | --- |
| `google-github-actions/auth@v3` | GCP **Workload Identity Federation** → `projects/609771777761/…/nightcrawler-github`, service account `nightcrawler-sheets@…`, `create_credentials_file: true` |
| Job `env` | `OPENAI_EMBEDDINGS_KEY`, `STAGING_DATABASE_HOST/PORT/USER/PASSWORD/DATABASE`, `DATABASE_PEM_CERT`, `STAGING_DATABASE_PEM_CERT` |

A hijacked `v3` tag mints live GCP credentials **inside a job already holding the staging database password and PEM cert**. `google-github-actions/auth` appears nowhere else in the repo, so this is its only pin anywhere.

### SHAs — resolved mechanically, not by hand

Every value came from the GitHub API at time of writing:

```
gh api repos/actions/checkout/git/ref/tags/v6            -> d23441a48e516b6c34aea4fa41551a30e30af803
gh api repos/google-github-actions/auth/git/ref/tags/v3  -> 7c6bc770dae815cd3e89ee6cdf493a5fab2cc093
gh api repos/oven-sh/setup-bun/git/ref/tags/v2           -> 0c5077e51419868618aeaa5fe8019c62421857d6
```

All three refs are of type `commit` (no annotated-tag dereference needed).

### Verified while here

I also mechanically checked every pin already proposed in #1022 — **all seven are correct**:

| Action | Pinned SHA | Verified |
| --- | --- | --- |
| `actions/checkout` | `df4cb1c0` | real commit, v6.0.3 |
| `actions/setup-node` | `24997072` | matches `tags/v6` |
| `actions/upload-artifact` | `043fb46d` | matches `tags/v7` |
| `oven-sh/setup-bun` | `0c5077e5` | matches `tags/v2` |
| `dependabot/fetch-metadata` | `21025c70` | matches `tags/v2` |
[...]

PR #1076, opened 2026-08-21
fix(ci): re-pin wait-on-check-action to v1.9.0 and normalise pin comments

Jose-Gael-Cruz-Lopez:
## Description

Stacked on #1022 — merge into `fix/pin-action-shas` before that PR lands.

### 1. The `wait-on-check-action` pin is a downgrade

`dependabot-auto-approve.yml:46` pinned the **v1.8.0** commit, but `main` already runs `@v1.9.0`. Merging as-is rolls the action back a minor version — and this is the hunk causing the conflict, so pre-resolving it here removes that conflict too.

Re-pinned to the v1.9.0 commit. Note `v1.9.0` is an **annotated** tag, so it needs dereferencing:

```
git/ref/tags/v1.9.0  -> fa4138d3…  (type: tag)
git/tags/fa4138d3    -> 2271c86c146b96545b4e871b855e10ffa6f50773  (commit)
```

Pinning `fa4138d3` directly would have been wrong — that is the tag object, not a commit.

### 2. Comment style was inconsistent within single files

`build.yml` used `# v6` on checkout but `#v6` on setup-node and setup-bun. Dependabot's `github-actions` updater parses that trailing comment to decide what to bump, so an unparseable one risks a pin silently never being updated. Normalised all to `# vX`.

## Your SHAs are correct — I verified all seven mechanically

This was the one property that could not be checked by eye, so I resolved each against the GitHub API:

| Action | Pinned | Result |
| --- | --- | --- |
| `actions/checkout` | `df4cb1c0` | real commit — **v6.0.3** (`v6` now points at `d23441a4`, so this is slightly behind but valid) |
| `actions/setup-node` | `24997072` | matches `tags/v6` |
| `actions/upload-artifact` | `043fb46d` | matches `tags/v7` |
| `oven-sh/setup-bun` | `0c5077e5` | matches `tags/v2` |
| `dependabot/fetch-metadata` | `21025c70` | matches `tags/v2` |
| `codecov/codecov-action` | `0fb71748` | matches `tags/v5` after annotated-tag deref |
| `lewagon/wait-on-check-action` | `eb1f8d8f` | real, but v1.8.0 — fixed here |

No SHA other than `wait-on-check-action` is changed by this PR.

## Correction to my earlier review comment

I previously wrote that this PR "skips" `sync-imps-to-staging.yml`. **That was unfair** — that file was added to `main` by `2f2060a` (#1026) *after* this branch was cut, so it was never in scope. Apologies.

The gap is real but independent, so I pinned it against `main` in #1075 (it is the only workflow using `google-github-actions/auth`, in a job holding the staging DB password and PEM cert).

## Suggested follow-up

Nothing in CI enforces pinning, so this can regress silently. Adding `zizmor` or `pin-github-action` to `code-quality.yml` would make it a standing gate rather than a one-time cleanup.

## Verification

All 11 workflow files parse as valid YAML. No logic, step ordering, `permissions` or `with:` values changed — only `uses:` refs and trailing comments.

## Checklist

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -16,12 +16,12 @@ jobs:
         uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6
 
       - name: Setup Node.js
-        uses: actions/setup-node@249970729cb0ef3589644e2896645e5dc5ba9c38 #v6
+        uses: actions/setup-node@249970729cb0ef3589644e2896645e5dc5ba9c38 # v6
         with:
           node-version: '20'
 
       - name: Setup Bun
-        uses: oven-sh/setup-bun@0c5077e51419868618aeaa5fe8019c62421857d6 #v2
+        uses: oven-sh/setup-bun@0c5077e51419868618aeaa5fe8019c62421857d6 # v2
         with:
           bun-version: 'latest'
 
@@ -38,7 +38,7 @@ jobs:
         run: bun run build-storybook
 
       - name: Upload Next.js build artifacts
-        uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a #v7
+        uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7
         if: always()
         with:
           name: nextjs-build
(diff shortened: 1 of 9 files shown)


vincenttodd:

Jose-Gael-Cruz-Lopez:
Closing: no longer needed. This fixed #1022's rollback of `wait-on-check-action` from v1.9.0 to v1.8.0, but #1022 is closed and #1084 landed the correct pin directly:

```
dependabot-auto-approve.yml:46
  uses: lewagon/wait-on-check-action@2271c86c146b96545b4e871b855e10ffa6f50773 # v1.9.0
```

Your note that v1.9.0 is an annotated tag needing dereferencing is exactly why that pin is right — worth having flagged.

PR #1077, opened 2026-08-21
fix(site): correct heading order, asset name and image sizes on /about

Jose-Gael-Cruz-Lopez:
## Description

Stacked on #1036 — merge into `fix/aboutpage-redesign` before that PR lands.

### 1. Heading order regressed to h1 → h3 → h2

`page-header.tsx:32-35` emits `<h1>{title}</h1>` immediately followed by `<h3>{subtitle}</h3>`. The new vision block then opens with `<h2>`. So the document outline **skips a level and then goes backwards** — before this PR the page was a clean h1 → h2.

A tagline is not a section heading, so the subtitle is now a `<p>` with all its typography classes intact. That also fixes the same pre-existing skip on **/research, /brand and the homepage**, which share this component.

`page-header.test.tsx` asserted `tagName === "H3"`; updated, with the reasoning recorded in the test.

### 2. Asset filename violates the kebab-case rule

`apps/site/public/marketing/aboutpagefamilyimage.jpg` → `about-family.jpg`, via `git mv` so history is preserved. AGENTS.md mandates kebab-case and every sibling complies (`who-we-are-header.webp`, `who-we-are-img.jpg`). Its single reference is updated.

### 3. `sizes` understated the slot, so Next served a too-small candidate

The image is `width={1440}` inside `max-w-[580px]`, rendering `w-full md:h-[580px] object-cover` — so at `md`+ the slot is up to **580 CSS px** (1160 device px at 2×). It declared:

```diff
-sizes="(min-width: 768px) 480px, 100vw"
+sizes="(min-width: 768px) 580px, 100vw"
```

At 480px the srcset picker chose one step too small and the photo softened on desktop.

## Needs a decision, not a patch — copy is duplicated on screen

`messages/about/en.json` sets both:

- `whoWeAre.subtitle` — "We believe sustainable agriculture is the foundation of a healthy planet and thriving communities."
- `whoWeAre.vision.description` — the same sentence, re-wrapped

They render **twice within one viewport** (PageHeader subtitle, then the vision paragraph), in both `en` and `es`. The PR's own test hedges around it:

```ts
expect(screen.getAllByText(/We believe sustainable agriculture is the foundation/i)).not.toHaveLength(0)
```

`getAllByText(...).not.toHaveLength(0)` instead of `getByText` is the tell that this is placeholder copy. Worth getting real vision copy before merging — not something I should invent.

## Three other things worth knowing

- **`md:text[17px]/[28px]` is a dead class** (`page-header.tsx:33`) — missing hyphen, should be `md:text-[17px]/[28px]`. So the intended `md`+ typography has never applied on any page using `PageHeader`. Left alone deliberately: fixing it changes rendering on four pages and belongs in its own PR. The test asserts the typo verbatim, so it keeps passing.
- **Do not delete `/marketing/who-we-are-img.jpg`.** The page stops using it, but `careers/fixtures/careers-landing-view.fixture.ts:56` still references it.
- The Careers CTA disappears from /about (`vision.cta` now points at `/research`). Careers stays reachable from the footer, but that is a content decision worth signing off rather than slipping through.

## Verification

- `tsc --noEmit`: clean
- Full site suite: **556 tests / 117 files pass**
- No i18n keys dropped — `intro.paragraphs` and `culture.*` had `about/page.tsx` as their only consumer, and en/es are updated symmetrically

## Checklist


vincenttodd:
is this doing the same thing as #1098?

PR #1078, opened 2026-08-21
fix(site): add popover data-slot, fix v4 origin syntax and the inert label

Jose-Gael-Cruz-Lopez:
## Description

Stacked on #1035 — merge into `feat/748-shadcn-calendar` before that PR lands.

The calendar swap itself is good — in particular the timezone round-trip (`toCalendarDate` / `fromCalendarDate` against `toDisplayDate`'s `timeZone: 'UTC'`) is **strictly better** than the `toISOString().split('T')[0]` + `valueAsDate` pairing it replaces. Three defects in the vendored files.

### 1. `popover.tsx` omits `data-slot`, so the calendar's own styling hook is dead

`calendar.tsx:34` styles itself with:

```
[[data-slot=card-content]_&]:bg-transparent [[data-slot=popover-content]_&]:bg-transparent
```

…but the vendored `PopoverContent` rendered no `data-slot` attribute. The selector never matched, so the calendar kept `bg-background` inside a `bg-popover` panel.

Added `data-slot` to `popover`, `popover-trigger` and `popover-content`, matching the convention already established in `dialog.tsx:13-14,41,63`.

**This is now covered by a test** that opens the picker and asserts the panel carries the attribute. I verified it genuinely catches the bug — removing the attribute again fails exactly that test and nothing else:

```
× marks the open panel with data-slot so the calendar renders transparent
  Tests  1 failed | 4 passed
```

### 2. Tailwind v3 syntax in a v4 project

```diff
-origin-[--radix-popover-content-transform-origin]
+origin-(--radix-popover-content-transform-origin)
```

This repo runs `tailwindcss ^4.2.2`. The bare `[--var]` shorthand is v3; v4 requires parentheses for a CSS-variable value. As written the utility emitted nothing, so the zoom/slide animations pivoted from the element centre instead of from the trigger. The rest of the vendored code already uses v4 syntax correctly (`h-(--cell-size)` in `calendar.tsx`), so this line was simply stale.

### 3. `<Label htmlFor>` pointed at a `<button>`, which is inert

A `<label for>` cannot label a button, so clicking the label did nothing — a regression against the native `<input type="date">` it replaced, which was focusable that way.

The code already compensated for screen readers with `aria-label={`${label}: ${displayValue}`}` on the trigger, which is correct. But the visible label kept a false association. It is now a plain `<span aria-hidden="true">`, so:

- no misleading `for` relationship
- the trigger remains the single source of the accessible name (no double announcement)

Also covered by a test asserting no `label[for="rotationYear"]` exists.

## Verification

Because this branch adds `@radix-ui/react-popover` and `react-day-picker`, I ran a real `bun install --frozen-lockfile` in an isolated worktree rather than relying on the root `node_modules` — otherwise every check is a false negative on unresolved modules.

- `tsc --noEmit`: **clean**
- Full site suite: **560 tests / 118 files pass** (2 new)
- New `data-slot` test confirmed to fail when the fix is reverted

## Not fixed here

- **`bun.lock` will conflict.** The lock hunk re-bumps `brace-expansion ^5.0.7 → ^5.0.8`, but `main` already pins `^5.0.8` at `package.json:80`. Regenerate the lock on top of current `main` rather than resolving by hand.
- **`DateField` is inline in the page-local form file.** It is generic (`control` + `name` + `label` + `disabled`) and the farm-profile and IMP forms will want it; feature-sliced design argues for its own module. Left alone to keep this diff reviewable.
- The test hand-rolls `globalThis.ResizeObserver ??= class {}` even though `resize-observer-polyfill` is already an `apps/site` devDependency. Worth moving into `vitest.setup.ts` once, since every future Radix-popover test needs it.

## Interaction with #1069

`management-zone-form.tsx` is also touched by #1069 (identity-column allowlist). #1035 keeps `location: [0, 0]` and does not address the coordinate clobber, so #1069's guard is still required after this lands. Whichever merges second needs a manual rebase on that file.

## Checklist

### DIFF
--- a/apps/site/src/app/(authenticated)/(no-sidebar)/account/(with-shell)/management-zones/[zone]/management-zone-form.test.tsx
+++ b/apps/site/src/app/(authenticated)/(no-sidebar)/account/(with-shell)/management-zones/[zone]/management-zone-form.test.tsx
@@ -76,6 +76,35 @@ describe('ManagementZoneForm date pickers', () => {
     );
   });
 
+  it('marks the open panel with data-slot so the calendar renders transparent', async () => {
+    const user = userEvent.setup();
+    render(<ManagementZoneForm zone={zone} canEdit />);
+
+    await user.click(screen.getByRole('button', { name: ROTATION_TRIGGER }));
+
+    // calendar.tsx styles itself with
+    // `[[data-slot=popover-content]_&]:bg-transparent`. Without this attribute
+    // on the panel the selector never matches and the calendar keeps
+    // `bg-background` inside a `bg-popover` panel — a silent visual bug no
+    // other assertion would catch.
+    const panel = document.querySelector('[data-slot="popover-content"]');
+    expect(panel).not.toBeNull();
+    expect(panel).toContainElement(
+      screen.getByRole('button', { name: /july 15th, 2026/i })
+    );
+  });
+
+  it('does not point a label at the trigger button', async () => {
+    render(<ManagementZoneForm zone={zone} canEdit />);
+
+    // A `<label for>` cannot name a button, so the association was inert and
+    // clicking the text did nothing. The trigger owns its accessible name.
+    expect(document.querySelector('label[for="rotationYear"]')).toBeNull();
+    expect(
+      screen.getByRole('button', { name: ROTATION_TRIGGER })
+    ).toBeInTheDocument();
+  });
+
   it('disables both pickers for read only accounts', () => {
     render(<ManagementZoneForm zone={zone} canEdit={false} />);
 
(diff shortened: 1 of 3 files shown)


vincenttodd:
could this be related refactoring #1096?

Jose-Gael-Cruz-Lopez:
Closing: structurally unmergeable, and its content is already in #1087.

`baseRefName` is `feat/748-shadcn-calendar` (#1035's branch), not `main`, and it is a draft — a stacked PR can't reach main on its own, and #1035 is now closed.

All three of its fixes are verified present in #1087: `data-slot` on popover/trigger/content (`popover.tsx:18,29,54`), the Tailwind v4 parenthesis syntax (`:58`), and the inert-label replacement (`form.tsx:108-109`). #1087 goes further with `outline-hidden` over `outline-none`.

The diagnosis in this PR was correct and is what made #1087 right — thank you for it.

PR #1079, opened 2026-08-21
fix(site): use react-vite storybook framework and finish the socials cut-over

Jose-Gael-Cruz-Lopez:
## Description

Stacked on #1018 — merge into `feat/786-socials-link-component` before that PR lands.

### 1. The storybook import is why Build and Code Quality fail

`social-links.stories.tsx:3` imported `@storybook/nextjs-vite`. `main` migrated to `@storybook/react-vite` in `bca54ad`; `nextjs-vite` is absent from `apps/site/package.json` and has **zero occurrences in `bun.lock`**.

`apps/site/tsconfig.json:25-33` includes `**/*.tsx`, so `tsc --noEmit` (`code-quality.yml:33`) and `next build` / `build-storybook` (`build.yml:32,35`) both resolve it and fail with TS2307. `eslint.config.mjs:49` ignores `*.stories.*`, so `lint` stays green — which is what makes the failure look odd.

### 2. Completing the cut-over — `not-found.tsx` was a third copy

The PR's stated goal is that social URLs, labels and icons "never duplicate them". But `not-found.tsx:20-41` still declared its own array of the same four links and rendered them at `:86-93`, so a URL change would still have needed two edits.

Migrated it to `<SocialLinks platforms={['instagram', 'linkedin', 'x', 'youtube']} />`, which also drops the now-unused `react-icons/fa6` import block. `SOCIAL_LINKS` is now genuinely the only place these URLs live.

The order and accessible labels are preserved exactly, so the existing 404 assertions are unaffected.

## Still needs the author — the marketing footer hunk no longer applies

I did not touch this, because it needs re-authoring rather than patching.

`main` restructured `footer.tsx`. The PR's pre-image expects `import Image from 'next/image'`, `socialMediaIcons` at ~line 52, and the socials in a standalone `<div className="flex flex-row flex-wrap gap-6">` at ~line 211. On `main`: no `next/image` import in that file, `socialMediaIcons` at 27-47, and the socials at ~139-145 inside `<div className="flex flex-wrap items-center gap-8 md:justify-end">` — **shared with the language switcher**. All three hunks fail.

That last point matters beyond the rebase: `SocialLinks` hard-codes `flex flex-row flex-wrap gap-6`, which is the `/go` footer's spacing. Dropping it into `main`'s bottom bar nests a `gap-6` row inside a `gap-8` row with nothing passed to compensate, so the marketing footer's spacing shifts. When re-authoring, either pass the existing classes through (`className="items-center gap-8"`) or make the wrapper class-free and let callers own layout.

## One behaviour change worth a decision

Marketing social links move from `<Link href>` (same tab) to `<a target="_blank" rel="noopener noreferrer">`. `rel` is correct, but the accessible names (`Visit our Instagram page`) do not announce the new tab. Either keep same-tab behaviour behind a prop, or append "(opens in a new tab)" to the labels.

## Worth keeping through the rebase

The `/go` Discord href had a trailing space (`'https://discord.gg/rFY3kc4deK '`). `SOCIAL_LINKS.discord.href` drops it — a genuine bug fix hiding in the refactor.

## Verification

- `tsc --noEmit`: clean
- Full site suite: **553 tests / 117 files pass**

## Checklist

### DIFF
--- a/apps/site/src/app/not-found.tsx
+++ b/apps/site/src/app/not-found.tsx
@@ -9,36 +9,7 @@ import { Link } from '@/i18n/config';
 import { createClient } from '@/lib/supabase/server';
 import { NextIntlClientProvider } from 'next-intl';
 import { getLocale, getMessages, getTranslations } from 'next-intl/server';
-import {
-  FaInstagram,
-  FaLinkedinIn,
-  FaXTwitter,
-  FaYoutube,
-} from 'react-icons/fa6';
-
-/** Social media links displayed at the bottom of the 404 page. */
-const socialMediaIcons = [
-  {
-    icon: <FaInstagram aria-hidden="true" />,
-    href: 'https://www.instagram.com/toddagriscience/',
-    ariaLabel: 'Visit our Instagram page',
-  },
-  {
-    icon: <FaLinkedinIn aria-hidden="true" />,
-    href: 'https://www.linkedin.com/company/toddagriscience/',
-    ariaLabel: 'Visit our LinkedIn page',
-  },
-  {
-    icon: <FaXTwitter aria-hidden="true" />,
-    href: 'https://x.com/toddagriscience',
-    ariaLabel: 'Visit our X (Twitter) page',
-  },
-  {
-    icon: <FaYoutube aria-hidden="true" />,
-    href: 'https://www.youtube.com/@toddagriscience',
-    ariaLabel: 'Visit our YouTube channel',
-  },
-];
+import SocialLinks from '@/components/common/social-links/social-links';
 
 /**
  * Async server component that fetches locale data and auth state.
@@ -83,17 +54,9 @@ async function NotFoundContent() {
               </div>
             </div>
             <div className="flex justify-center">
-              <div className="flex flex-row flex-wrap gap-6">
-                {socialMediaIcons.map((val) => (
-                  <Link
-                    key={val.href}
-                    href={val.href}
-                    aria-label={val.ariaLabel}
-                  >
-                    {val.icon}
-                  </Link>
-                ))}
-              </div>
+              <SocialLinks
+                platforms={['instagram', 'linkedin', 'x', 'youtube']}
+              />
             </div>
           </div>
         </FadeIn>
--- a/apps/site/src/components/common/social-links/social-links.stories.tsx
+++ b/apps/site/src/components/common/social-links/social-links.stories.tsx
@@ -1,6 +1,6 @@
 // Copyright © Todd Agriscience, Inc. All rights reserved.
 
-import type { Meta, StoryObj } from '@storybook/nextjs-vite';
+import type { Meta, StoryObj } from '@storybook/react-vite';
 import SocialLinks from './social-links';
 
 const meta: Meta<typeof SocialLinks> = {


Jose-Gael-Cruz-Lopez:
## Why CI is red on this PR, and why the change is still correct

Being upfront: the `Build Application` and `Code Quality Checks` failures here are caused by my own storybook import change, in a way that resolves itself only after the base branch is rebased. Detail so nobody wastes time on it.

The failure is:

```
error TS2307: Cannot find module '@storybook/react-vite'
```

That is the **inverse** of the failure on the base branch. The reason is dependency resolution:

| Context | `package.json` / `bun.lock` has | Correct import |
| --- | --- | --- |
| This base branch today | `@storybook/nextjs-vite` | `nextjs-vite` |
| `main` (since `bca54ad`) | `@storybook/react-vite` | `react-vite` |

This PR merges into the base branch, whose lockfile predates the migration — so `react-vite` genuinely is not installed there yet, and CI fails.

**After the base branch merges to `main` the situation reverses.** The base branch does not modify `package.json`, so `main`'s version wins on merge: `react-vite` is present and `nextjs-vite` is gone. At that point `nextjs-vite` is the broken import and `react-vite` is the correct one.

### So the ordering is

1. Rebase the base branch onto current `main` (it needs this regardless — it is stale, and its other story files that still name `nextjs-vite` resolve from `main` on merge)
2. Merge this PR

After step 1, CI here goes green and the import is correct for the destination.

### If you would rather not carry a red PR

Take only the non-storybook half of this branch and change the import yourself during the rebase. That half is independent and useful on its own. I would rather flag the tradeoff than quietly leave an import that breaks the moment the branch lands on `main`.

### Separately: `Security Audit` is failing everywhere right now

Not related to this PR. `GHSA-r292-9mhp-454m` (node-tar, affects `<=7.5.20`) was published after `main`'s last green run, and `main`'s `overrides` pinned `tar: ^7.5.19` which resolved to exactly `7.5.20`. I opened #1082 against `main` to bump it — `bun audit --audit-level=high` exits 0 with that change.

Stale branches will still fail Security Audit after #1082 for a *different* reason: branches cut before `bca54ad` (#1064) predate the `overrides` block remediating the nanoid/undici advisories, so their merge refs resolve an old tree. That one needs a rebase, not a dependency change.

vincenttodd:
@Jose-Gael-Cruz-Lopez explain this to a 1 y/o

akeight:
We're removing storybook, so that blocker can be removed

Jose-Gael-Cruz-Lopez:
Closing: superseded by #1085, and this branch's base is dead.

`baseRefName` is `feat/786-socials-link-component` — the head of #1018, which was **closed** on 2026-08-23 — so this can't reach `main` on its own.

Both of its unique changes are already in #1085, verified blob-by-blob:
- `social-links.stories.tsx:3` — #1085 already imports `@storybook/react-vite`.
- The duplicate four-link array in `not-found.tsx` — #1085 removes it **and** adds `not-found.test.tsx` coverage, which this branch omits.

All three of your dependency claims check out against main: `package.json:98` pins `@storybook/react-vite`, `.storybook/main.ts` sets that framework, and `nextjs-vite` has 0 occurrences in `bun.lock`. That was a real breakage worth flagging.

#1085 also adds `social-profiles.ts` as an icon-free constants layer, which fixes a live drift this branch doesn't touch: `(unauthenticated)/layout.tsx:37` and `not-found.tsx:28` currently disagree on the LinkedIn URL by a trailing slash. Your own description notes the marketing footer hunk needs re-authoring — #1085 already did that against current main.

PR #1080, opened 2026-08-21
fix(site): drop obsolete hunks and restore the account heading and home link

Jose-Gael-Cruz-Lopez:
## Description

Stacked on #1012 — merge into `fix/970-move-farm-contact-to-left-menu` before that PR lands.

The #970 idea is sound and still needed: `account-side-menu.tsx` on `main` shows no farm or contact info, and the layout still renders `AccountHeader`. Four problems in the current branch.

### 1. `get-embedding.ts` would break CI on every PR and every release

The branch re-adds a module-scope `throw`. `main` deliberately replaced that with a lazy `embeddingsClient()` in #1065, and its JSDoc records why. The throw fires at **import**:

```
pr-database-compatibility-check.yml:94   →  bunx tsx scripts/migrate-remote-db.ts --env staging
release-database-compatibility-check.yml:92 →  … --env prod
migrate-remote-db.ts:21                  →  import { getArg } from './lib/importer-lib'
importer-lib.ts:12                       →  import { getEmbedding } from '.../get-embedding'
```

`OPENAI_EMBEDDINGS_KEY` appears **once** across all workflows (`sync-imps-to-staging.yml:19`); neither compatibility job has it. Reverted to the merge-base version so `main`'s lazy client wins on merge.

Note the import at `importer-lib.ts:12` cannot just be deleted — `importer-lib.ts:149` really calls `getEmbedding`. Deferring the check into the function is the only shape that works, which is what `main` does.

### 2. Three files were already on `main` verbatim

Pure conflict surface, no remaining value. All reverted to merge-base so `main`'s copies apply cleanly:

| File | Already on main |
| --- | --- |
| `proxy/auth.ts` | `:8-14` (`/account` in `protectedUrls`), `:137-139` (exact-match branch) |
| `proxy/auth.test.ts` | `:62-64` |
| `packages/db/scripts/seed-local-db.ts` | `:154` (dynamic import) |

That takes the PR from 9 files to 5, all of them actual #970 work.

### 3. Removing `AccountHeader` deleted the only `<h1>` — and the only way out

`account-header.tsx:34` held the **only `<h1>` in the entire account subtree**, and `:16` the only `role="banner"`. The replacement was `<p>{farmName}</p>`, so every `/account/*` page became heading-less.

It also removed the wordmark and the `← Home` link. I checked every layout above `account/(with-shell)` — **none renders any chrome** (`(shell)/(widgets)` and `(shell)/(misc)` do, but they are sibling route groups). So the account area had no route back to the app except the browser back button.

Rather than resurrect the header — your no-header direction looks deliberate — the side menu now owns both:

- `farmName` renders as `<h1>`, restoring a single top-level heading
- a `← Home` link sits above it, restoring the escape route

### 4. Contacts were unlinked, and the component had no JSDoc

Email and phone rendered as bare `<p>` text in what is explicitly a contact panel. Now `mailto:` / `tel:` links, guarded against the `'Not set'` sentinel from `account/util.ts:14` so the placeholder does not become `mailto:Not set`. Phone strips formatting for the `tel:` href while displaying the formatted value.

`account-side-menu.tsx:20` had a `//` line comment on an exported component; AGENTS.md requires JSDoc. Replaced with a full block including `@param`/`@returns`.

Also deleted `account-header.tsx`, which had zero remaining callers (AGENTS.md clean cutover). Recoverable from history if you decide to restore the header instead.

## Verification

- `tsc --noEmit`: clean
- Full site suite run serially: **511 tests / 110 files pass**

A note on that: running in parallel on my machine produced intermittent timeouts in unrelated files (`signup`, `contact`, `login`, `submit-button`). I confirmed these are load flakes, not regressions — `submit-button` passes in isolation both with and without my changes, and the whole suite is green with `--no-file-parallelism`.

## Still needed before this can merge

**The branch must be rebased onto `(no-sidebar)/account/`.** `main` collapsed `(shell)/(accounts)/account/**` to `(no-sidebar)/account/**` in #1005, so all five remaining files are rename/modify conflicts. This PR does not fix that — it only removes the four hunks that should not be rebased at all, so there is less to move.

Also relevant: #1021 claims the same ticket but contains no farm/contact change. Recommend closing it in favour of this branch.

## Checklist


vincenttodd:
could be related refactoring #1096

Jose-Gael-Cruz-Lopez:
Closing: superseded by #1096, and this branch cannot be rebased cheaply.

`baseRefName` is `fix/970-move-farm-contact-to-left-menu` — the head of #1012, which was **closed** on 2026-08-23 — so it can't reach `main` alone.

More decisively, all six files here live under `(shell)/(accounts)/account/**`, a route group `main` removed in #1005. Everything is now `(no-sidebar)/account/**`. `merge-tree` against main gives 4 conflicts including a rename/delete on `account-header.tsx` that no tool can resolve automatically. Your own description says as much: *"The branch must be rebased onto `(no-sidebar)/account/`… This PR does not fix that."*

The three hunks you correctly identified as not-to-be-carried are all honoured in #1096 — the `get-embedding.ts` module-scope throw (which would break both DB-compatibility workflows), plus `proxy/auth.ts` and `seed-local-db.ts` which had already landed on main. That analysis is what made #1096 correct.

PR #1081, opened 2026-08-21
fix(site): key the rate limiter on a trustworthy client IP

Jose-Gael-Cruz-Lopez:
## Description

Stacked on #960 — merge into `fix/936-rate-limiting` before that PR lands.

The store choice in #960 is right and worth saying plainly: `Ratelimit` over `Redis.fromEnv()` with a sliding window means counters live in Upstash, so **this survives multi-instance serverless deployment**. No module-global `Map`. Two defects in how the key is derived, plus a docs fix.

### 1. The limiter keyed on the client-controlled end of `x-forwarded-for`

`get-client-ip.ts` took `forwardedFor.split(',')[0]` — the **left-most** entry.

`x-forwarded-for` is append-style: each proxy appends the address it received the connection from, so the left-most element is whatever the *client* sent. An attacker rotating `x-forwarded-for: <random>` gets a fresh Redis bucket on every request, and the limiter becomes a no-op.

Now, in order:

1. `x-vercel-forwarded-for` — set by the Vercel edge, not client-writable
2. the **right-most** `x-forwarded-for` hop — the one our own proxy appended
3. `x-real-ip`

### 2. All header-less traffic shared one global bucket

`getClientIp` returned the literal string `'unknown'`, passed straight to `check(ip)`. Every request lacking the header worldwide therefore shared `site:public-write:unknown` — **5 requests per minute in total, globally**. That is a self-inflicted denial of service, not a defence.

It now returns `null`, and `enforceRateLimit` skips the check with a `logger.warn` rather than keying on a placeholder. Skipping matches the limiter's existing deliberate fail-open stance, and on Vercel a forwarding header is always present, so this path is effectively unreachable in production.

### 3. `env.example` described the Upstash vars as "Optional"

Combined with `getRedis()` returning `null` on missing config, a production deploy without them yields **no rate limiting at all**, silently, with one warn per cold start. Reworded to "REQUIRED IN PRODUCTION", making clear that unset is for local dev / CI / preview only.

## Verification

`@upstash/ratelimit` and `@upstash/redis` are added by this branch, so I ran a real `bun install --frozen-lockfile` in an isolated worktree — otherwise every check is a false negative on unresolved modules.

- `tsc --noEmit`: clean
- Full site suite, run serially: **461 tests / 103 files pass**
- `get-client-ip` tests rewritten for the new contract, including a case asserting a spoofed left-most entry is ignored:

```ts
'x-forwarded-for': 'evil-1, evil-2, evil-3, 203.0.113.7'  →  '203.0.113.7'
```

## Not addressed here — still worth doing before this ships

**Coverage.** This guards `submitPublicInquiry` and `submitFormSubmission` (which does cover careers and platform-access applications). Still unguarded:

| Path | Risk |
| --- | --- |
| `signup/actions.ts:211` `resendApprovedApplicantActivationEmail` | Unauthenticated, **sends a real email** per call. Loopable → email-bombing an applicant and burning the Supabase invite quota. Highest-severity gap. |
| `signup/actions.ts:295` `signUp` | Unauthenticated; writes `farm` + `user` rows and creates a Supabase auth user, behind a token whose `applicationId` is a sequential integer — brute-forceable with no limit. |
| `api/revalidate/route.ts:45` | Non-constant-time secret compare, no limit; success purges the whole site cache. |
| `auth/confirm/route.ts:14` | Unauthenticated OTP verification, unlimited attempts. |

Correctly excluded: `api/stripe/webhook/route.ts:30-50` verifies the Stripe signature first.

**Fail-open on Redis errors.** `getRedis()` returning `null` and the `catch` around `limiter.limit` both return `{ success: true }`, so a missing config and an outage are indistinguishable from being under the limit. Consider throwing at module init in production when the vars are absent.

**Log level.** `enforceRateLimit` → `throwActionError` → `lib/utils/actions.ts:63` `logger.error`. Being rate limited is expected, so under attack each blocked request becomes an error-level line. `warn` would be better.

## Rebase note

Two of the eleven files patch `(contact)/contact/actions.ts` and its test, both **deleted on `main`** by the Sanity form-builder migration (`submitPublicInquiry` has zero occurrences repo-wide). Those hunks are a modify/delete conflict — drop them. The `forms/[slug]/actions.ts` call site still applies and already covers contact via `DynamicForm`.

## Checklist

### DIFF
--- a/apps/site/env.example
+++ b/apps/site/env.example
@@ -98,8 +98,10 @@ NEW_RELIC_APP_NAME="TODD_SITE_TESTING"
 NEW_RELIC_LICENSE_KEY="secret"
 
 # Upstash Redis for IP-based rate limiting on public write endpoints (contact
-# form, CMS forms). Optional — when unset, rate limiting fails open (allows
-# requests) so local dev / CI / preview keep working. Provision via the Vercel
+# form, CMS forms). REQUIRED IN PRODUCTION. When unset, rate limiting fails
+# open (allows every request) with a single warn per cold start, so a
+# production deploy without these has no rate limiting at all. Unset is
+# intended only for local dev / CI / preview. Provision via the Vercel
 # Marketplace Upstash integration, which injects these automatically.
 UPSTASH_REDIS_REST_URL="https://your-db.upstash.io"
 UPSTASH_REDIS_REST_TOKEN="your-token"
--- a/apps/site/src/lib/rate-limit.ts
+++ b/apps/site/src/lib/rate-limit.ts
@@ -148,6 +148,12 @@ export const publicWriteRateLimit = createRateLimiter({
  * limiter, and throws an action error when the limit is exceeded. This is the
  * one-line guard server actions should call before doing any work.
  *
+ * When no trustworthy IP is available the check is skipped rather than keyed on
+ * a shared placeholder, which would put every such request into a single global
+ * bucket and rate-limit the whole world to `requests` per window collectively.
+ * On Vercel the edge always sets a forwarding header, so this is effectively
+ * unreachable in production and matches the limiter's existing fail-open stance.
+ *
  * @param check - Limiter to enforce. Defaults to {@link publicWriteRateLimit}.
  * @param message - Error surfaced to the client when rate limited.
  */
@@ -156,6 +162,11 @@ export async function enforceRateLimit(
   message = 'Too many requests. Please try again shortly.'
 ): Promise<void> {
   const ip = getClientIp(await headers());
+  if (!ip) {
+    logger.warn('[rate-limit] No trustworthy client IP; skipping check.');
+    return;
+  }
+
   const { success } = await check(ip);
   if (!success) {
     throwActionError(message);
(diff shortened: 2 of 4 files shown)


vincenttodd:

Jose-Gael-Cruz-Lopez:
Closing: your fix is already in #1086, which can actually merge.

`baseRefName` is `fix/936-rate-limiting` (#960's branch), not `main`, so this can only update #960 — and #960 is now closed as unmergeable against main.

Your core finding was correct and important: `forwardedFor.split(',')[0]` takes the left-most, client-controlled hop, so header rotation made the limiter a no-op. #1086 implements the same trust order and goes slightly further — it factors a shared `lastHop()` helper and applies it to **all three** headers, including `x-vercel-forwarded-for`, which this branch only `.trim()`s and would therefore mis-key if it ever arrived comma-listed.

This branch also still carries the `(contact)/contact/actions.ts` hunks for files deleted from main, and its Security Audit check is failing. Thanks for catching the bypass.

PR #1082, opened 2026-08-21
fix(deps): bump tar override past GHSA-r292-9mhp-454m

Jose-Gael-Cruz-Lopez:
## Description

**Security Audit is currently red on `main`, which blocks every open PR in the queue.** This is a 3-line fix for it.

### Cause

`GHSA-r292-9mhp-454m` — node-tar, uncontrolled recursion in `mapHas`/`filesFilter` allowing an uncatchable stack-overflow DoS via a crafted long-path tar with member selection. It affects **`<=7.5.20`** and was published *after* main's last green Security Audit run (`a688cfe`, 2026-08-21T02:38).

`package.json` already carried `"tar": "^7.5.19"` in `overrides`, but the lockfile had resolved that to exactly **7.5.20** — inside the vulnerable range. So the override existed and was still vulnerable.

```
tar@7.5.20
  workspace:@nightcrawler/sanity > sanity > @sanity/cli > tar
  high: node-tar: Uncontrolled recursion in mapHas/filesFilter … (<=7.5.20)
```

### Fix

`overrides.tar: ^7.5.19 → ^7.5.22`, lockfile regenerated. Resolves to `tar@7.5.22`.

### Verification

`.github/workflows/security.yml:26` is a single assertion — `bun audit --audit-level=high`. Run locally against this branch:

```
$ bun audit --audit-level=high
bun audit v1.3.14
$ echo $?
0
```

Diff is 3 lines total (1 in `package.json`, 2 in `bun.lock`).

### Risk

`tar` reaches the tree only via `sanity > @sanity/cli > tar` — a build/CLI path, not application runtime. Patch-level bump within the same major, so no API change.

## How I found it

While opening stacked fixes across the open PRs, **every one** of them failed Security Audit, including one that targets `main` and changes nothing but a workflow YAML file. That isolated the cause to the dependency tree rather than to any PR's changes.

Worth noting for triage: this makes the audit signal on the whole queue currently meaningless. Once this lands, stale PRs will still fail Security Audit for a *different* reason — branches cut before `bca54ad` (#1064) predate the `overrides` block that remediates the nanoid/undici advisories, so their merge refs resolve an old tree containing `@storybook/nextjs-vite` and friends. Those need a rebase, not a dependency change.

## Checklist


rcjasub (approved):
LGTM

PR #1083, opened 2026-08-22
fix(account): correct the principal operator key and guard a missing farm row

Jose-Gael-Cruz-Lopez:
## Description

Supersedes #973, rebased onto current `main` with review fixes folded in. **0 merge conflicts.**

### What the original PR fixed
Two live bugs in `apps/site/src/app/(authenticated)/(no-sidebar)/account/db.ts`:
1. The loader emitted `name` for the principal operator while `(with-shell)/users/page.tsx` reads `principalOperator.firstName` — so the **Name row rendered blank in production**.
2. `getAccountFarmData` destructured the first query row with no guard → `TypeError`/500 when no farm row exists.

The original PR conflicted only because `main` collapsed `(shell)/(accounts)/account/**` into `(no-sidebar)/account/**`. Neither bug had been fixed on `main`.

### Fixes added after audit
An auditor **disproved** the original claim that the added test defended the fix. Mutating `firstName` back to `name` left **all tests green and `tsc` exit 0** — because `users/page.test.tsx` does `vi.mock('../../db', ...)`, so `db.ts` was executed by *zero* tests repo-wide.

- **New `db.test.ts`** (4 tests) that actually exercises the loader.
- **Root cause closed:** `principalContact` was an unannotated object literal, so TypeScript's excess-property check never fired. Annotating it makes the typo a **compile error**.
- **New `AccountContact` type** — `Partial<UserSelect>` was a lie, since `firstName` holds a full display name (`toDisplayName(first, last)`), not the schema column. Non-`Partial`, so a dropped key is also now a compile error.
- Guard uses `notFound()` (page data loader convention here) rather than a raw throw.

### Verification
`tsc --noEmit` clean · account suite **7 files / 12 tests** pass · `merge-tree` **0 conflicts**

Mutation-proven both directions: reverting the key turns the new test **red** (`expected undefined to be 'Alex Owner'`) and emits `TS2353`.

## Checklist
- [x] Tests included
- [x] Inline docs included
- [x] Accessible
- [x] Conventional commits


vincenttodd:

vincenttodd (approved):

PR #1086, opened 2026-08-22
feat(site): rate-limit public write endpoints (#936)

### LINKED ISSUE #936: No rate limiting on public write endpoints

Jose-Gael-Cruz-Lopez:
## Description

Supersedes #960 (issue #936), rebased onto current `main` with review fixes. **0 merge conflicts.**

Rate-limits public, unauthenticated write endpoints with an Upstash Redis sliding window.

### The store choice was right — kept
Counters live in Upstash, shared across lambdas, so this **survives multi-instance serverless**. No module-global `Map`.

### Why the original conflicted
Two of its eleven files patched `(contact)/contact/actions.ts` and its test — **both deleted on `main`** by the Sanity form-builder migration (`submitPublicInquiry` has zero occurrences repo-wide). Those hunks are dropped; contact coverage is retained via the guarded `forms/[slug]` action, which is the live path.

### Security fixes folded in

**1. Header-rotation bypass.** The original keyed on `forwardedFor.split(',')[0]` — the **left-most** entry. `x-forwarded-for` is append-style, so that element is whatever the *client* sent; rotating it minted a fresh bucket per request and made the limiter a no-op. Now: `x-vercel-forwarded-for` (edge-set, not client-writable) → **right-most** hop → `x-real-ip`.

**2. Self-inflicted DoS.** Missing headers returned the literal string `'unknown'`, used directly as the Redis key — so every header-less request worldwide shared one bucket, collectively capped at 5/min. Now returns `null` and the check is skipped with a `logger.warn`, consistent with the module's deliberate fail-open stance.

**3. Silent prod misconfiguration.** `env.example` called the Upstash vars "Optional". Since `getRedis()` returns `null` when absent, a production deploy without them has **no rate limiting at all**. Reworded to required-in-production.

### Coverage extended beyond the original
| Endpoint | Preset | Why |
| --- | --- | --- |
| `resendApprovedApplicantActivationEmail` | 3 / 10 min | Unauthenticated and **sends a real email** per call — loopable into email-bombing and burning Supabase invite quota |
| `signUp` | 10 / 10 min | Unauthenticated; writes farm+user rows and creates an auth user, gated only by a token whose `applicationId` is a sequential integer |

`api/stripe/webhook` deliberately **not** limited — it verifies the Stripe signature before doing any work.

### Audit findings addressed
- The hand-written `bun.lock` entries were independently verified **byte-exact against the npm registry**, and a real `bun install --frozen-lockfile` succeeds. Kept.
- A test titled "configures distinct namespaces" never read `prefix` — provably vacuous. Setting two presets to the same prefix passed 15/15 on the old test; the rewritten one goes red.

### Verification
`tsc` clean (deps overlay) · full suite **634 tests / 127 files** · `merge-tree` **0 conflicts**

## Checklist
- [x] Tests included
- [x] Inline docs included
- [x] Accessible
- [x] Conventional commits


vincenttodd:

akeight (approved):
LGTM

PR #1087, opened 2026-08-22
feat(site): replace account date inputs with the shadcn calendar

### LINKED ISSUE #748: Swap calendar for ShadCN calendar

Jose-Gael-Cruz-Lopez:
## Description

Closes #748 
Supersedes #1035, rebased onto current `main` with review fixes. **0 merge conflicts.**

Replaces the two native `<input type="date">` fields on the management-zone form with a shadcn calendar in a popover.

### Timezone round-trip kept exactly
`toCalendarDate`/`fromCalendarDate` paired with `toDisplayDate`'s `timeZone: 'UTC'` is **strictly better** than the `toISOString().split('T')[0]` + `valueAsDate` pairing it replaces. Deliberately not simplified.

### Fixes folded in

**1. `data-slot` was missing, so the calendar's own styling hook was dead.** `calendar.tsx` styles itself with `[[data-slot=popover-content]_&]:bg-transparent`, which never matched — the calendar kept `bg-background` inside a `bg-popover` panel. Now emitted on root/trigger/content, matching the convention in `dialog.tsx`.

**2. Tailwind v3 syntax in a v4 project.** `origin-[--radix-popover-content-transform-origin]` emits nothing under `tailwindcss ^4.2.2`; a CSS-variable value needs parentheses. Verified by compiling both forms with the repo's own toolchain — the bracket form produced `transform-origin: --radix-...` with no `var()` wrapper.

**3. `<Label htmlFor>` pointed at a `<button>`.** A `<label for>` cannot name a button, so the association was inert and clicking did nothing — a regression against the native input. Now a non-interactive `<span aria-hidden="true">`; the trigger already carries the accessible name via `aria-label`.

**4. `outline-none` → `outline-hidden`.** Compiled both with tailwindcss 4.2.2: only `outline-hidden` emits the `@media (forced-colors: active)` rule that preserves an outline in Windows High Contrast.

### Audit finding: deleted tests restored
The rebase kept the source fixes but **silently deleted the two regression tests defending them** (file shrank 116 → 87 lines). Restored and mutation-proven: removing `data-slot` turns one red; re-adding a `<Label htmlFor>` turns the other red.

`forwardRef` → function component was independently verified **not** to break ref threading (React 19 passes `ref` as a prop).

### Requires a human step
`bun.lock` is deliberately **untouched** — a hand-edited lockfile would be worse than none. Run `bun install` after merge to record `@radix-ui/react-popover ^1.1.23` and `react-day-picker ^10.0.1`.

### Verification
`tsc` **0 errors** with deps resolved (9 baseline errors are all unresolved-module artifacts) · full suite **613 tests / 125 files** · `merge-tree` **0 conflicts**

## Checklist
- [x] Tests included
- [x] Inline docs included
- [x] Accessible
- [x] Conventional commits


vincenttodd:
could be related to a refactoring on #1096, if not, probably good

akeight:
@vincent, this is seperate from 1096, this adds a nicer looking calender in place of another. I'll check this visually when I can get a new dev login account to see it.

akeight (approved):
LGTM

PR #1088, opened 2026-08-22
fix(site): make logout navigate instead of rejecting

### LINKED ISSUE #848: [Bug]: Chunk error after logging out and/or navigating to `/news` page

Jose-Gael-Cruz-Lopez:
## Description
Closes #848 
Supersedes #949, rebased onto current `main`. **0 merge conflicts.**

### The bug
`logout()` in `apps/site/src/lib/auth-client.ts` — a `'use client'` module — called `redirect('/')` from `next/navigation`. That is a **server-only** API, so it throws `NEXT_REDIRECT` and `logout()` **rejects instead of returning**.

The repo already knew: `logout-link.tsx` carried the comment *"If logout doesn't redirect automatically, manually redirect"*.

The un-workarounded call sites are genuinely broken today — `(authenticated)/error.tsx:34` and `auth-error-fallback.tsx:35` call `logout()` fire-and-forget with no `.catch`, so **users cannot log out from the very screen telling them to log out**.

### The fix
`window.location.replace('/')`, plus a **clean cutover**: the three callers that worked around this (`logout-link.tsx`, `account-side-menu.tsx`, `desktop-gate.tsx`) had their `router.push('/')` removed, along with the now-unused `useRouter` imports and the stale comment. Otherwise a client push would race a document navigation.

A hard navigation is genuinely the right call after `signOut()` — it drops the RSC payload cache and all client state.

### Deliberately dropped from the original
- **The news-page hunk** removed a `'use server'` from `(marketing)/news/page.tsx`. `main` rewrote that page to render `ArticleIndex` and the line is already gone. This was the sole conflict.
- **The chunk-reload hunk** in `app/error.tsx` performed `sessionStorage` side effects **during render** (unsafe — StrictMode double-renders, React may discard renders) and its session flag was never cleared on success, so the guard only ever worked once per session.

> **Scope note for review:** this is half of what #949 originally proposed. The logout fix is sound and lands here; the chunk-error handling needs rework as a `useEffect` in its own PR. Flagging explicitly rather than quietly shipping a reduced scope.

Also note there is **no `global-error.tsx`** anywhere in `apps/site/src/app`, so root-layout chunk failures stay uncaught regardless — another reason that half deserves separate treatment.

The original PR title (`Title: fix(site): ...`) violates `commitlint.config.cjs` `type-enum`; corrected here.

### Verification
`tsc` clean · auth + logout + account suites pass · `merge-tree` **0 conflicts**

## Checklist
- [x] Tests included
- [x] Inline docs included
- [x] Accessible
- [x] Conventional commits


vincenttodd:

akeight (approved):
LGTM

PR #1090, opened 2026-08-22
chore: remove dead code and unused dependencies, add knip

Jose-Gael-Cruz-Lopez:
## Description

**Part 2 of 3** splitting #1011. **0 merge conflicts.** Land this **first**.

Removes dead code and unused dependency declarations, and adds `knip` so dead code stays measurable.

### Contents
- **20 unreferenced modules deleted**
- **15 dependency declarations removed** — `apps/site` (11 deps + 1 devDep), `apps/internal` (4), `packages/db` (1)
- `knip.json` + a `bun run knip` script
- Stops leaking `ImpNotesFormValues` as an ambient global — now exported and imported explicitly

### Every deletion and removal was re-derived independently
Not taken on trust. For each deleted file, every exported symbol was grepped repo-wide, then every deleted module specifier, then both barrel files (`components/common/index.ts`, `components/ui/index.ts`) were confirmed to re-export none of them.

**Result: zero deletions break a live import.** Two worth calling out:
- `(shell)/(misc)/accept/actions.ts` + `types.ts` — the real consumer is `(unauthenticated)/accept/components/accept-form.tsx`, which imports a *different* file of the same name.
- `(authenticated)/components/apply-button.tsx` — its only reference sits inside a block comment, and the relative path already resolved to a nonexistent module.

### Deliberately excluded: `share-article.tsx`
The original PR deleted it. **Do not** — `main` modified it in `23b891c` (#1020: repaired the LinkedIn share URL, added UTM params) and added a 92-line test that did not exist at the original merge base. It still has no external importers, so it is arguably dead, but deleting it now discards #1020's work. Worth a separate decision.

### Restored after audit
An auditor found the split had silently dropped the original PR's deletion of the orphaned `components/go/header/` pair. Re-derived on current `main`: zero matches for `@/components/go`, no dynamic imports, and the live component is `app/(go)/creators/components/header/header.tsx`. Confirmed orphaned and deleted.

> Note: the shipped `knip.json` is **blind** to that class of defect — knip's vitest plugin registers `**/*.test.tsx` as entry points, so a file reachable only from its own sibling test looks used. Removing test entries is insufficient because the plugin re-adds them, and disabling vitest would flag hundreds of legitimate test-only files. Recorded as a known limitation rather than papered over.

### Requires a human step
`bun.lock` deliberately untouched. Run `bun install` after merge.

### Verification
Full CI on the combined three-branch merge: `turbo type-check` **4/4 workspaces** · `format:check` clean · **608 tests / 124 files** · lint clean · Storybook build · **Next 16 production build** — all green.

## Checklist
- [x] Tests included
- [x] Inline docs included
- [x] Accessible
- [x] Conventional commits


vincenttodd:

akeight (approved):
LGTM!

PR #1091, opened 2026-08-22
chore: rename files to kebab-case and update importers

Jose-Gael-Cruz-Lopez:
## Description

**Part 3 of 3** splitting #1011. **0 merge conflicts.** Land **second**: `1011b` → `1011c` → `1011a`.

Renames files to kebab-case per AGENTS.md, and updates every importer.

### Safety checks performed
- **Zero case-only renames** — re-confirmed independently. Every rename inserts a hyphen, so macOS's case-insensitive filesystem is not a hazard.
- **Every old module specifier has zero remaining references** after the rewrites.
- **i18n directory renames are safe** — verified, not assumed. The translation namespace lives *inside* the JSON (`messages/cookiePreferences/en.json` has key `"cookiePreferences"`), so renaming the directory leaves `useTranslations('cookiePreferences')` working untouched.
- `header/types/header.ts` → `header/types.ts` is safe because that directory holds only that one file.

### Dropped as obsolete
The original's `apps/site/tsconfig.json` hunk — `main` already generalised the Storybook include to `.storybook/**/*.ts{,x}` globs.

### One correction to the record
An earlier report claimed `share-article.tsx` was untouched by all three branches. **That was wrong** — this branch modifies it and its test, by exactly **one line each**:

```
share-article.tsx:11       '@/lib/hooks/useCurrentUrl' -> '@/lib/hooks/use-current-url'
share-article.test.tsx:9   vi.mock(...) same path update
```

That is the required importer rewrite following the `useCurrentUrl.tsx` → `use-current-url.tsx` rename in this same branch — legitimate and necessary; omitting it would break the import.

**Regression-checked:** filtering out that single line, both files are byte-identical to `main`. #1020's work is untouched — `taggedShareUrl`, the `utm_source`/`utm_medium`/`utm_campaign` params and the LinkedIn share-offsite URL are all unchanged.

### Merge-order note
This is the branch that invalidates other open PRs' paths, so land it in a quiet window and announce it.

### Verification
Order-independence proven: `main → a → b → c` and `main → c → b → a` produce the **identical tree** (`efa8fb5`). Full CI on the combined merge — type-check ×4 workspaces, 608 tests, lint, Storybook build, Next 16 production build — all green.

## Checklist
- [x] Tests included
- [x] Inline docs included
- [x] Accessible
- [x] Conventional commits


vincenttodd:

andrewdack:
These are still camelCase even though same-directory files got converted:

      apps/site/.storybook/decorators/withNextNavigation.tsx
      apps/site/src/lib/hooks/useMediaQuery.tsx
      apps/site/src/lib/hooks/useMediaQuery.test.tsx

  So `apps/site/.storybook/preview.ts` imports `with-fonts` and
  `with-storybook-provider` from kebab paths and `withNextNavigation` from camel,
  on adjacent lines. Same in `src/lib/hooks/` next to `use-current-url.tsx`,
  `use-window-width.tsx` and `use-cookie-preferences.ts`.

 #1011 deleted all three of the camel case files as dead code, so skipping them here
  made sense. But #1090 doesn't carry those deletions, because
  `ResponsibilitiesSection` still renders at
  `apps/site/src/app/(unauthenticated)/[locale]/(marketing)/about/page.tsx:95`,
  and it and `competencies-section` both use `useMediaQuery`. Deleting them would
  have cut a section off the About page.

  With the deletions gone, I think we should rename these files to kebab-case in this PR. Other than that the changes look valid

akeight (approved):
LGTM!

PR #1092, opened 2026-08-22
fix(db): verify the TLS certificate chain on staging and prod

Jose-Gael-Cruz-Lopez:
## Description
Supersedes #1061. **0 conflicts.** Closes a live TLS hole — and fixes a way the original PR *reopened* it.

### The hole on main
`packages/db/src/schema/connection.ts:14-19` does `ssl: !process.env.NODE_TLS_REJECT_UNAUTHORIZED ? {ca} : false`. Setting that var to the **secure** value `"1"` turns SSL off entirely. `drizzle-prod.config.ts:17` and `drizzle-staging.config.ts:19-24` both ship `rejectUnauthorized: false`, and `db-target.ts` neutered hostname checks with `checkServerIdentity: () => undefined`. **Staging and prod accept any certificate today.**

### Why not just merge #1061
Its `declaredPlaintextHosts()` does not subtract the managed clusters, so `DATABASE_PLAINTEXT_HOSTS=<prod host>` yields `ssl:false` for prod — the same hole, one env var away. Its own `env.example` admitted it. Proven with a side-by-side probe: the PR returns `ssl === false` for the prod URL; this branch returns `{rejectUnauthorized: true}`.

### What this adds on top
- `managedClusterHosts()` subtraction so the escape hatch cannot name a real cluster
- `caCertSourceForConnection()` — a missing staging cert now names `STAGING_DATABASE_PEM_CERT`, not the wrong var
- A named error for an unparseable `DATABASE_URL` (a raw `#` or `/` in the password defeated `new URL()` and failed closed with a confusing message)
- `turbo.json` `globalEnv` entries. **Load-bearing:** Turborepo 2.x defaults to `envMode: "strict"`, which *strips* undeclared vars from task environments — so without this the cert-var fix could never see `STAGING_DATABASE_HOST` under `turbo dev`
- Both DB-compatibility workflows raised `verify-ca` → `verify-full`. Proven-safe parity, not a gamble: `db-reconcile.yml:116,250,308` already runs verify-full against the same clusters with the same certs. Local dumps stay `sslmode=disable`.

### Verification
42/42 `db-ssl` tests · `tsc` clean · grep-proven: no path reaches `rejectUnauthorized: false` or a neutered `checkServerIdentity`.

## Checklist


vincenttodd:

vincenttodd (approved):

PR #1093, opened 2026-08-22
fix(site): stop management zone updates from writing identity columns

Jose-Gael-Cruz-Lopez:
## Description
Supersedes #1069. **0 conflicts.** Fixes a tenant-scoped mass assignment **and** a live data-loss bug the original missed.

### Security fix
`management-zones/[zone]/actions.ts` does `.set({ ...input })`. The edit form seeds `defaultValues` from the whole row, so every submit posts `id`/`farmId`/`createdAt`/`updatedAt` back. `.where()` constrains **which** row, never **what** is written — and `id` is a writable `serial`, not `GENERATED ALWAYS`. So a farm Admin could move their zone to another farm. An allowlist closes it.

### Why not just merge #1069
Two reasons. Its base is `fix/vitest-localstorage-shim` (#1068), **not `main`** — so it targets the wrong branch. And it misses an adjacent bug.

### The extra bug this fixes
`management-zone-form.tsx:35` seeds `location: [0, 0]`. The lat/long inputs carry no `defaultValue`, so **every save silently overwrites the zone's real coordinates with (0,0)**. Now `zone.location ?? [0,0]`.

### Also fixed
react-hook-form 7.72.1 `getFieldValueAs` maps `''` → `NaN` under `valueAsNumber` and → `new Date('')` under `valueAsDate`. So clearing a coordinate stored the literal `(NaN,NaN)` (Postgres accepts float8 NaN) and clearing a date threw `RangeError` inside `PgDate.mapToDriverValue`. Now `setValueAs` yields `undefined` (leave alone) or `null` (clear), the action drops `location` unless both coordinates are finite, and rejects an Invalid Date by name.

A bare `null` is left as-is — verified via drizzle `.toSQL()` that it correctly emits a NULL parameter.

### Verification
20/20 tests off `main` — confirmed it does **not** depend on #1068's shim.

## Checklist


vincenttodd:

vincenttodd (approved):

PR #1094, opened 2026-08-22
feat(site): add back navigation to account subpages

Jose-Gael-Cruz-Lopez:
## Description
Supersedes #1033. **0 conflicts.**

Adds `backHref`/`backLabel` to the shared `AccountInfo` wrapper so account subpages get consistent back navigation. Uses `next/link`, not `useRouter().back()` — a plain server-rendered anchor, so no SSR break and no dead end on a direct deep link.

### Why not just merge #1033
`account-info.stories.tsx:3` imports `@storybook/nextjs-vite`, which has **0 entries in `bun.lock`** — `.storybook/main.ts:45` pins `@storybook/react-vite`. It only appears to work from a stale local `node_modules`; a fresh CI install fails, and `apps/site/tsconfig.json` includes `**/*.tsx` so `tsc` and `next build` break while `lint` stays green.

**Its green checks are stale** — they last ran 2026-07-29, before the storybook migration landed 2026-08-11. Do not trust them.

### Fixes on top
- Story imports `@storybook/react-vite`
- Removes the now-duplicate `Back to zones` link inside `management-zone-form.tsx` — `AccountInfo`'s `backHref` replaces it, so the page had two links to one destination
- Destination-specific `backLabel` on every call site instead of six identical "Back" links, which are indistinguishable in a screen-reader link list
- Dead `backLabel?.trim()` optional chain simplified (the prop is defaulted, so it could never fire)
- JSDoc on all three `AccountInfo` exports

### Verification
8/8 test files · `tsc` clean · duplicate-link removal is mutation-proven.

> **Merge note:** shares one test file with #1093 (add/add). Trivial union — keep #1093's file, append this branch's `describe` block. Validated green.

## Checklist


vincenttodd:
same thing as #1096 comment, should maybe make into a refactor and address changes on a call

vincenttodd (approved):

PR #1095, opened 2026-08-22
fix(site): restore careers disclosures with an accommodation inquiry link

Jose-Gael-Cruz-Lopez:
## Description
Supersedes #1067. **0 conflicts.**

Teaches `Disclaimer` to render next-intl rich-text link tags and rewrites `careers.disclaimers` 2/3/4 in **both** locales to use `<inquiry>`/`<coverage>`/`<privacy>` instead of a raw support email and a bare URL — keeping the address off a scrapeable public page.

### Settled: the /go videos
`creatorsbig.mp4` and `creatorssmall.mp4` (referenced at `(go)/creators/page.tsx:29,36`) are present on `main`, on #1067's branch, and here — all at **identical blob hashes**. My earlier #1071 landed. Nothing to restore.

### Fixes on top
- **Accessibility:** outbound disclaimer links now carry a lucide `ExternalLink` glyph plus a `title` announcing a new tab. `title` becomes the accessible *description*, deliberately not the name — the tests assert exact accessible names, so sr-only text or `aria-label` would have broken them.
- Reverts #1067's unrelated `.husky/commit-msg` change. `npx --no` is a never-install contract that fails loudly; `bunx` fetches from the registry, turning a fail-closed hook fail-open. That churn does not belong in a careers-copy PR. *(It did add useful `"$1"` quoting — trivially re-addable separately.)*
- Adds a 77-line rich-text test.

### One thing reviewers should know
The global `vitest.setup.intl.js` rewrite is **required** — the old mock did `enMessages[namespace]`, which cannot resolve a dotted namespace like `careers.disclaimers`, so those lookups silently rendered `[careers.disclaimers.0]` placeholders.

Blast radius measured: exactly 3 dotted namespaces / 13 lookups flip from placeholder to real copy repo-wide. **Full suite green — no test depended on the placeholder output.**

### Verification
25/25 tests · both `en` and `es` updated symmetrically, no leftover raw email or bare URL.

## Checklist


vincenttodd:
@Jose-Gael-Cruz-Lopez I feel horrible but this is actually already fixed, disclosures were moved to job listings

vincenttodd:
closing bc this is fixed

PR #1096, opened 2026-08-22
feat(site): move farm contact details into the account side menu

Jose-Gael-Cruz-Lopez:
## Description
Supersedes #1012 (ticket #970). **0 conflicts.**

Moves the farm name and primary contact details out of the account page header and into the account left-hand menu.

### Honouring the maintainer directive
@vincenttodd commented on #1012:
> *"Removing the old account header (the Todd workmark is, but i don't think the home or farm name are part of the header)"*

So: wordmark **is** part of the header; Home link and farm name are **not**. This branch implements exactly that — `AccountHeader` is now wordmark + `role="banner"` only.

The Home link moves into the side menu inside its **own** `<nav aria-label="Site">`, sibling to the existing `<nav aria-label="Account sections">`. A `<nav>` is itself a landmark, so it stays in the landmark list rather than hiding inside the `<aside>`. Two escape routes survive: the wordmark links to `/`, plus the new Home link. Deleting the header outright — as the original did — would have removed the only `role="banner"` **and** the only route out of `/account/*`, since no layout above `account/(with-shell)` renders any chrome.

### Heading structure fixed
The earlier rebase left the only `<h1>` inside `<aside>` while `<main>` had **zero** headings. Now `AccountInfo`'s **existing** section title is promoted h2 → h1 (rather than adding a competing heading), `AccountInfoSection` h3 → h2 so levels don't skip, and the side-menu farm name becomes an `<h2>` that names the `<aside>` via `aria-labelledby`.

Composed-render proof across all shell routes: exactly one `<h1>`, parent landmark `MAIN`, exactly one banner.

### Also
Contacts are real `mailto:`/`tel:` links via new `toMailtoAddress`/`toTelNumber` helpers that only linkify genuinely mailable/dialable values — so `mailto:Not set`, `tel:` on `"N/A"`, and empty strings all render as plain text.

Three hunks from the original are correctly **not** carried: the `get-embedding.ts` module-scope throw (would break both DB-compatibility workflows), and `proxy/auth.ts` + `seed-local-db.ts`, which already landed on `main`.

## Checklist


vincenttodd:
@Jose-Gael-Cruz-Lopez Not sure, maybe make this into a refactor issue and I can circle back this week, cuz this might be a old issue

PR #1097, opened 2026-08-22
fix(site): restore jsdom web storage when node shadows it in tests

Jose-Gael-Cruz-Lopez:
## Description
Supersedes #1068. **0 conflicts.** Low priority — merge last.

Node 22+ ships a global `localStorage`/`sessionStorage` that shadows jsdom's, so `localStorage.clear()` in `beforeEach`/`afterEach` throws and aborts the hook chain on dev machines.

### Scope is genuinely inert in CI
`.nvmrc` is `20.20.1`, and `build.yml:21` / `code-quality.yml:21` / `testing.yml:21` all pin `node-version: '20'`. Node 20 has no global Web Storage, so the shim **never fires in CI**. This is a local-dev-on-Node-22 convenience, nothing more.

It also does **not** gate #1093 (management zone), contrary to earlier belief — that branch's 20 tests pass off `main` without this shim.

### Improvement over #1068
The original buried 84 lines of untested logic in `vitest.setup.ts`. This extracts it into `apps/site/src/test/memory-storage.ts` with a 116-line test file (9 tests).

The returned object is now **Proxy-wrapped** so bracket and dot access and `delete` behave like real `Storage` — the original supported only the method surface, so `localStorage.foo = 'x'` would have written a plain own-property invisible to `getItem`/`length`/`key`. It also does not clobber a healthy jsdom Storage, and does not leak state between test files.

## Checklist


vincenttodd:

akeight (approved):
LGTM!

PR #1098, opened 2026-08-22
fix(site): about page redesign with heading order and asset fixes

Jose-Gael-Cruz-Lopez:
## Description
Supersedes #1036. **0 conflicts.** ⚠️ **DO NOT MERGE YET — blocked on marketing copy.**

### Why #1036 itself must not merge
It is **semantically stale in a way git cannot see**. Cut at `d6731e1`, before `responsibilities-section` landed. Merging it onto `main` **breaks 4 tests** (proven), because its `messages/about/{en,es}.json` drop the entire `whoWeAre.responsibilities` key set and its `page.tsx` stops rendering the section — orphaning a live component. Git reports it MERGEABLE.

This branch restores the responsibilities keys and renders the section.

### Heading regression fixed
`page-header.tsx:33` emits `<h3>` for the subtitle directly under the `<h1>`, and the new vision block then opens `<h2>` — so the outline goes **h1 → h3 → h2**, skipping a level and then going backwards. The subtitle is now a `<p>`, which also fixes the same pre-existing skip on **/research, /brand and the homepage**. Verified safe: no test anywhere asserts an h3 from `PageHeader`. A "never skips a heading level" test is added.

### Also
- Asset renamed `aboutpagefamilyimage.jpg` → `about-family.jpg` via `git mv` (kebab-case), zero dangling references
- `sizes` corrected 480px → 580px; the slot is `max-w-[580px]`, so the srcset picker was choosing one candidate too small and the photo softened on desktop

### 🛑 The blocker — needs marketing, not engineering
`whoWeAre.subtitle` and `whoWeAre.vision.description` are the **same sentence**, rendered twice in one viewport, in both `en` and `es`. The PR's own test hedges around it with `getAllByText(...).not.toHaveLength(0)` instead of `getByText` — which is the tell that this is placeholder copy.

I will not invent product copy. Someone needs to write distinct vision copy before this ships.

### Verification
23/23 tests · `tsc` clean.

### Noted, deliberately not fixed
`md:text[17px]/[28px]` in `page-header.tsx` is a **dead class** (missing hyphen), so the intended `md`+ typography has never applied on any page using it. Fixing it changes rendering on four pages and belongs in its own PR.

## Checklist


vincenttodd:
@Jose-Gael-Cruz-Lopez whats the problem with this pr and building off of Aram's about header change? Theres these changes as well:

- (Our research ->) button doesn't look like its either using the right component or is hard coded (it should match the one on like the homepage)
- the "We are creating generative farms, but will also consider our mission fulfilled if our work aids others to achieve this outcome." should be regular weight, size it right tho. I also want to change "generative" to "sustainable" to stay consistent
- want to change the text below "our vision" text to "We believe agriculture should remain a human centric career that heals the planet and supports communities."
- could we bring in the image and that text a bit more like the figma mockup?

akeight (approved):
LGTM!

PR #1099, opened 2026-08-22
feat(site): add a cursor follower to the newsroom article rows

### LINKED ISSUE #340: Cursor Animation for News Page
Implement a cursor animation for when it is hovered over a news article on /news like on https://www.skky.com/investments

Kenyang1:
## Description

Fixes #340

Implements the custom cursor animation requested for `/news`, modeled on the reference in the issue (skky.com/investments): a dark circle that trails the pointer and expands with a localized label — **View** / **Ver** — while hovering an article row. On the reference site each hover target carries `data-cursor-hover` / `data-ctext` attributes read by a follower element; this PR uses the same pattern.

### How it works

- **`components/common/cursor-follower/`** — a new client component. It renders only on devices that can hover (`matchMedia('(hover: hover) and (pointer: fine)')` via the existing `useMediaQuery` hook), so touch devices and the server render nothing. It listens for pointer events at the **document** level and resolves hover targets with `closest('[data-cursor-label]')`, so any server-rendered markup can opt in just by carrying the attribute — no per-element handlers. Movement uses framer-motion `useMotionValue` + `useSpring` for the trailing feel; `useReducedMotion` drops the springs entirely. The bubble is purely decorative: `aria-hidden`, `pointer-events-none`, `fixed`, `z-[100]`.
- **`ArticleIndex`** (shared by `/news` and `/research/index`) gains an optional `cursorLabel` prop. When set, article rows are tagged with `data-cursor-label` and the follower mounts. Only `news/page.tsx` passes it (`t('cursor.view')`); the research index intentionally keeps the native pointer.
- **`globals.css`** hides the native cursor over tagged rows *only* while the follower is actually mounted (`html[data-cursor-follower]` is set/removed by the component) and only on fine-pointer devices — so the pointer can never vanish on touch, before hydration, or on pages without the follower.
- **Translations**: `articleIndex.cursor.view` added to `en` and `es`.

### A subtle RSC bug worth knowing about

The first working version exported the `data-cursor-label` attribute name as a constant **from the `'use client'` component file** and used it in the server component as a computed object key. Every jsdom test passed — but in the real app the attribute silently vanished from the SSR HTML. Reason: when a server component imports from a `'use client'` module, *every* export (even a string constant) arrives as a **client-reference proxy**, not the value; React DOM then drops the garbage attribute name. The constants now live in a directive-free `constants.ts` imported by both sides, with a comment explaining why. Reviewers: this is why the row tagging and the component live in separate modules.

### Testing

- `cursor-follower.test.tsx`: renders nothing without a fine pointer; shows/clears the hovered label (real `pointerover`/`pointerout` dispatch); hides when the pointer leaves the window; removes its `<html>` marker on unmount.
- `article-index.test.tsx`: rows tagged + follower mounted when `cursorLabel` is set; untouched when omitted (research).
- `news/page.test.tsx`: end-to-end through the page — rows carry the localized "View" label. jsdom has no `matchMedia`, so a shared stub (`src/test/stub-match-media.ts`, mirroring the one in `useMediaQuery.test.tsx`) is installed file-wide.
- Storybook: `Common/CursorFollower` with news-style demo rows and a custom-styled variant.
- Full suite: 615/615. Verified in the dev server: SSR carries the attribute on all 18 rows in both locales (`View`/`Ver`), zero on `/research/index`, label toggling and `cursor: none` confirmed live.

### Notes for reviewers

- The issue predates the `/news` rebuild onto `ArticleIndex`; implementing it there means the research index automatically stays unaffected. If design wants the follower elsewhere later, it's one prop.
- The label text is chosen by the page (already localized), so the component itself needs no i18n dependency.

## Checklist


vincenttodd:
just throwing this here from figma (this size is same as pill button or slightly smaller)

Screenshot 2026-08-22 at 11 16 03 PM

Jose-Gael-Cruz-Lopez:
@vincenttodd thanks — the Figma reference was the right call, the first pass was a circle with "View" and didn't match. Pushed `cbaa949` to align it.

**Shape** — `size-20` circle → pill. `rounded-full` with `h-[42px]`, `px-5`, `gap-2` and auto width, so it sizes to its label.

**Size** — you said *"same as pill button or slightly smaller"*, so I anchored it to the real thing rather than eyeballing the screenshot. `components/common/button/themes/button.tsx` is the pill source of truth (`rounded-full`, `gap-2`, `px-5 py-2` at `lg`), and `page-header.tsx` pins that to `h-[47px]` in practice. The follower is `h-[42px]` with `text-base` and a `size-6` arrow, against the Button's `size-8`. So: same geometry and same design language, one step down. The intent is that it reads as a *cursor*, not as a control someone should try to click — it's `pointer-events-none`, so a click-looking pill that can't be clicked would be a bit of a lie.

**Arrow** — same `HiArrowLongRight` the Button uses, so it's the identical glyph rather than a lookalike. Marked `aria-hidden` — the whole bubble is already `aria-hidden` and decorative, and I didn't want the arrow leaking into any accessible name.

**Copy** — `View`/`Ver` → **Read**/**Leer**, matching the Figma. Renamed the message key `articleIndex.cursor.view` → `.read` in both catalogs and updated the one consumer in `news/page.tsx`.

Also added a regression test pinning the pill geometry and the decorative arrow, and mutation-tested it: restoring the old circle turns exactly that one test red (`1 failed | 6 passed`) and nothing else — so it actually defends the design instead of just passing.

Full `apps/site` suite: **125 files / 618 tests** green. I also merged current `main` in to pick up the `tar` advisory fix (#1082), so Security Audit is no longer red on this branch.

Two things worth a look from you since they're judgement calls, not spec:

1. **`h-[42px]` vs the Button's `47px`** — that's my read of "slightly smaller". Happy to take it to 44 or straight to 47 if you'd rather it match exactly.
2. **Spanish "Leer"** — the direct equivalent of "Read". If the brand voice would prefer "Ver más" or similar, say the word.

Preview: https://nightcrawler-git-feat-340-news-cursor-animation-todd-marketing.vercel.app/news — hover any article row.

vincenttodd (approved):
whatever tests need fixed otherwise frontend lgtm

PR #1106, opened 2026-08-24
refactor(site): restyle the auth error page to match the site

### LINKED ISSUE #1101: refactor auth error page
### Problem Statement

current authentication error page is not consistent with updated frontend style

old ref:

Image

### Context

Frontend like font weight/size needs updated to match the rest of the site. 

MAKE SURE the button is a component used elsewhere, bc this button should look exactly like other outline styles on the marketing site

Image

### Constraints

_No response_

### Technical Details

_No response_

### User Story

_No response_

### Use Case

_No response_

Orlando275:
Closes #1101

## Screenshots

Rendered through the Storybook story added in this PR
(`Authenticated/AuthErrorFallback`), so the page can be seen without having to
trigger a real authentication failure.

**Desktop — 1440x900**

pr1106-escritorio

**Mobile — 390x844**

pr1106-movil

The "old ref" screenshot of what this replaces is on the issue.

## What

The authenticated error page kept a one-off button: it declared `variant="outline"` and
then overrode it with inline classes into a solid dark block — the opposite of an outline.
Its copy also sat at heading sizes the rest of the site no longer uses.

It now uses the shared marketing `Button` (`@/components/common/button`), the same one
`not-found.tsx` and the about page use, so the outline styling comes from a single source
instead of being hand-rolled.

## Matching the mockup

Following the design on the issue and the links section from the design comment:

- one line of copy at body size, no `<br />`
- the "If you continue to experience issues…" paragraph is gone — it is absent from both
  target mockups
- copy reads "logout" / "Logout", matching the mockup and the existing `LogoutLink` label
- Terms · Privacy · System Status near the bottom, separated by `|`

## Shared button

The marketing `Button` only rendered a `Link`, so an action such as logging out could not
reuse it. It now renders a real `button` element when given an `onClick`, and is otherwise
untouched — every existing usage and its tests are unchanged.

Internal links use `next/link` (as the rest of the route group does); only the status page,
which lives on another host, stays a plain anchor.

## Error boundary

`(authenticated)/error.tsx` held a near-identical copy of the same page. It now delegates
to `AuthErrorFallback`, so the two authentication failure paths cannot drift apart.

## Notes for review

- **No next-intl here.** The entire `(authenticated)` group has zero files using
  translations, so the copy stays hardcoded to match the group. Glad to change this if the
  platform is meant to be localised.
- **The TODD wordmark stays** in the header. The old ref shows it; the new mockups do not
  clearly show that corner. Easy to drop if the design removed it on purpose.
- **Out of scope:** the other three `logout()` callers await the result and redirect to
  `/`. This page does not, so the viewer stays put after logging out. That is pre-existing
  behaviour and is not touched here.

## Verification

`bun validate` passes: type-check, lint (0 errors), 614 site + 13 db tests, build.

A Storybook story was added for the page (`Authenticated/AuthErrorFallback`), so it can be
reviewed without having to trigger a real authentication failure.


vincenttodd (approved):

akeight:
I was wondering what happened to this, lol. I was in the middle of reviewing it, haha.

Issue #1115, opened 2026-08-26
bun.lock has a duplicate key, so bun silently ignores it and installs are not reproducible


Jose-Gael-Cruz-Lopez:
## Description

`bun.lock` contains two byte-identical `"knip/zod"` entries (currently lines 2759 and 2761). Bun's lockfile parser rejects the second:

```
InvalidPackageKey: failed to parse lockfile: 'bun.lock' at bun.lock:2761:5
```

## Impact

`bun install` does not treat a parse failure as fatal. It warns and continues:

```
warn: Ignoring lockfile
```

It then re-resolves **every** dependency from the semver ranges in `package.json`. CI runs `bun i`, so each run installs whatever satisfies those ranges at that moment rather than the pinned versions. Consequences:

- Installs are not reproducible; CI and local machines can silently diverge.
- Version pins in `bun.lock` provide no protection.
- Dependency upgrades land with no diff and no review.
- `bun install --frozen-lockfile` cannot be used at all — it fails outright.

Observed drift on `main` today (pinned → actually installed):

| Package | Pinned in bun.lock | Installed by `bun i` |
| --- | --- | --- |
| `zod` | 4.3.6 | 4.4.3 |
| `@hookform/resolvers` | 5.2.2 | 5.9.1 |
| `react-hook-form` | 7.72.1 | 7.86.0 |

This is not theoretical: the zod drift is what broke `type-check` on `main` with no source change (see the companion issue).

## Origin

Introduced by #1090, the commit that added knip — most likely a lockfile merge resolution that duplicated the entry.

## Expected

`bun.lock` parses, `bun install --frozen-lockfile` succeeds, and the pinned versions are installed.

## Follow-up worth considering

CI uses `bun i`, which will keep masking this class of problem. Switching the install step to `bun install --frozen-lockfile` would make a corrupt or stale lockfile fail loudly instead of silently drifting. Raised separately so it can be decided on its own.

Jose-Gael-Cruz-Lopez:
The duplicate `"knip/zod"` key was removed by #1113. Verified on current main: the key appears once and `bun install --frozen-lockfile` now succeeds, so the pins are honored again.

Narrowing this issue to the remaining hardening item rather than closing it outright:

[...]

PR #1116, opened 2026-08-26
fix(site): type the phone schema input as string for the form resolver

### LINKED ISSUE #1114: type-check fails on main: phone schema input types as unknown
## Description

`Code Quality Checks` fails on `main` at the **Run type checking** step, with no source change to the file involved:

```
src/app/(unauthenticated)/accept/components/accept-form.tsx(36,5): error TS2322:
Type 'Resolver<{ ...; phone: unknown; ... }, any, {...}>' is not assignable to
type 'Resolver<{ ...; phone: string; ... }, any, {...}>'.
  Type 'unknown' is not assignable to type 'string'.
```

## Cause

`userInfo.phone` in `apps/site/src/lib/zod-schemas/onboarding.ts` is built with `z.preprocess()`:

```ts
phone: z.preprocess((phone: string) => { ... }, z.e164('Invalid phone number')),
```

Zod types `preprocess` as accepting `unknown` regardless of the callback's parameter annotation. That `unknown` propagates into `z.input<typeof acceptInviteSchema>`, while `AcceptInvite` is `z.infer<>` — the **output** type, where `phone` is `string`.

`zodResolver` is typed against the schema's *input* type, so in `accept-form.tsx`:

```ts
useForm<AcceptInvite>({ resolver: zodResolver(acceptInviteSchema) })
```

the resolver's `phone: unknown` cannot satisfy the form's `phone: string`.

## Why it started without a code change

Confirmed version-dependent. `apps/site` declares `zod: "^4.3.6"`; the lockfile pins `4.3.6`, but `^4.3.6` resolves to `4.4.3` today, and 4.4.x types `z.preprocess()` differently:

| Dependencies | Code | Result |
| --- | --- | --- |
| zod 4.4.3 / @hookform/resolvers 5.9.1 | current `preprocess` | fails (the CI error) |
| zod 4.3.6 / @hookform/resolvers 5.2.2 | current `preprocess` | passes |

CI installs the newer versions because the lockfile is being ignored — see the companion issue on the duplicate `bun.lock` key.

## Expected

`bun run type-check` passes, and the phone schema types its input as `string` so it is not sensitive to the zod patch range.

## Secondary problem

Because `preprocess` accepts `unknown`, a non-string value reaches the callback and throws a `TypeError` on `.replaceAll` instead of failing validation cleanly.

Jose-Gael-Cruz-Lopez:
## Description

Fixes #1114

`type-check` is failing on `main` at a file nobody changed:

```
src/app/(unauthenticated)/accept/components/accept-form.tsx(36,5): error TS2322:
Type 'Resolver<{ ...; phone: unknown; ... }>' is not assignable to
type 'Resolver<{ ...; phone: string; ... }>'.
  Type 'unknown' is not assignable to type 'string'.
```

`userInfo.phone` was built with `z.preprocess()`, which zod types as accepting `unknown` no matter what the callback's parameter is annotated as. That `unknown` propagates into `z.input<typeof acceptInviteSchema>`, while `AcceptInvite` is `z.infer<>` — the output type, where `phone` is `string`. `zodResolver` is typed on the *input* type, so it could not satisfy `useForm<AcceptInvite>`.

The fix swaps `z.preprocess(fn, z.e164())` for `z.string().transform(fn).pipe(z.e164())`. The normalization body is unchanged; what changes is that the input type is now pinned at `string`, so input and output agree and the resolver lines up with the form type.

Leading with `z.string()` also closes a latent runtime hole. Under `preprocess`, a non-string value reached the callback and threw a `TypeError` on `.replaceAll` instead of failing validation. It now reports `expected string, received number`.

### Why this is the right layer to fix it

The breakage is version-sensitive — `zod: "^4.3.6"` resolves to `4.4.3` today, and 4.4.x types `preprocess` differently:

| Dependencies | Code | Result |
| --- | --- | --- |
| zod 4.4.3 / resolvers 5.9.1 (what CI installs) | `preprocess` | fails — the CI error |
| zod 4.4.3 / resolvers 5.9.1 | this PR | passes |
| zod 4.3.6 / resolvers 5.2.2 | `preprocess` | passes |

The underlying reason CI installs 4.4.3 at all is a corrupt `bun.lock`, handled separately in #1115. That one restores the pins; this one makes the schema correct under **both** versions, so it does not re-break when zod is legitimately upgraded. They are complementary — this PR is the one that unblocks `main`.

## Verification

- `bun run validate` exits 0 — format, `type-check` across all 4 packages, lint, 742 tests, build.
- Normalization confirmed unchanged against `555-123-4567`, `5551234567`, `15551234567`, `1-555-123-4567`, `+15551234567`, `+447911123456` — all still produce the same E.164 output — with `123`, `""` and `abc` still rejected.
- Verified passing under both the pinned and the drifted dependency sets.

## Checklist

<sub>Existing coverage exercises this path (742 tests pass); the change is a type-level correction to a shared schema with behavior held constant, and the reasoning is documented inline above the field. No component or accessibility surface is touched.</sub>

https://claude.ai/code/session_013mgA7GjEeuiQ1jBSgdUGth

### DIFF
--- a/apps/site/src/lib/zod-schemas/onboarding.ts
+++ b/apps/site/src/lib/zod-schemas/onboarding.ts
@@ -8,25 +8,32 @@ export const userInfo = z.object({
   lastName: z.string().min(1, 'Last name is required').max(100),
   farmName: z.string().min(1, 'Farm name is required').max(100),
   email: z.email('Invalid email address').max(100),
-  // Very basic phone number handling
-  phone: z.preprocess((phone: string) => {
-    const withoutDashes = phone.replaceAll('-', '');
-    const digits = withoutDashes.replace(/\D/g, '');
+  // Very basic phone number handling. Uses `.transform().pipe()` rather than
+  // `z.preprocess()` because preprocess types its input as `unknown`, which
+  // leaks into `z.input<>` and makes `zodResolver` incompatible with a form
+  // typed on `z.infer<>`. Leading with `z.string()` also turns a non-string
+  // into a validation error instead of a TypeError inside the callback.
+  phone: z
+    .string()
+    .transform((phone) => {
+      const withoutDashes = phone.replaceAll('-', '');
+      const digits = withoutDashes.replace(/\D/g, '');
 
-    if (withoutDashes.startsWith('+')) {
-      return withoutDashes;
-    }
+      if (withoutDashes.startsWith('+')) {
+        return withoutDashes;
+      }
 
-    if (digits.length === 10) {
-      return `+1${digits}`;
-    }
+      if (digits.length === 10) {
+        return `+1${digits}`;
+      }
 
-    if (digits.length === 11 && digits.startsWith('1')) {
-      return `+${digits}`;
-    }
+      if (digits.length === 11 && digits.startsWith('1')) {
+        return `+${digits}`;
+      }
 
-    return phone;
-  }, z.e164('Invalid phone number')),
+      return phone;
+    })
+    .pipe(z.e164('Invalid phone number')),
   // Yes, this is stupid. See: https://github.com/colinhacks/zod/discussions/2801
   website: z.string().url().or(z.literal('')).optional(),
   instagramHandle: z.string().max(100).optional(),


Jose-Gael-Cruz-Lopez:
Superseded by #1113, which landed the same fix while this was in review.

That hotfix arrived at the identical root cause and effectively the same shape — `z.string().overwrite(fn).pipe(z.e164())` versus the `z.string().transform(fn).pipe(z.e164())` here. Both pin the schema's input type at `string` so `zodResolver` stays assignable to `useForm<AcceptInvite>`, and both keep the normalization identical. #1113 also added `onboarding.test.ts` covering the E.164 cases, which is a genuine improvement over this PR.

`type-check` is green on main. Closing as redundant; no action needed.

PR #1118, opened 2026-08-26
refactor(site): adopt the shared closing cta across marketing pages

### LINKED ISSUE #1112: refactor bottom of page component
### Problem Statement

current BoP are outdated and don't match newer site styles

Image

Image

### Context

use figma for styles on most of the marketing site pages. header size is 48px, button text is 14px.

Image

there is a current updated look on career, however its not component 

Image

### Constraints

_No response_

### Technical Details

_No response_

### User Story

_No response_

### Use Case

_No response_

Orlando275:
## Description

Fixes #1112

### Heads-up: the component already existed

The issue says the careers closing band "is not component". It already is —
`MarketingCenteredPageCta` has been in `marketing-blocks` since the careers
rebuild, and its docblock describes it as the bottom-of-page stack. What was
missing is that **only `/careers` used it**. So this PR adopts it on the two
pages still running the old markup rather than building anything new.

### What changed

**`/about` and `/research`** dropped their heading-as-link + `HiArrowLongRight`
blocks in favour of the shared component. Both now render an `<h2>` with a
separate pill link underneath, matching careers.

**Heading size** went from `text-3xl md:text-[34px]` to `text-4xl
md:text-[48px]`, per the 48px in the issue's Figma reference. Note this also
changes the careers closing band, which shipped at 34px — flagging it explicitly
since it wasn't part of the original ask.

**Arrows.** Removed the icon from both pages, and stripped the eight literal `→`
characters that were living inside `messages/careers/{en,es}.json` — including
`"View careers →"`, which fed the very component being adopted.

**New copy needed.** The old pattern had one string doing double duty as heading
and link text; the new one needs a heading *and* a button label. I've proposed:

| Page | Heading (existing) | Button (new) |
| --- | --- | --- |
| `/about` | What we do · Qué hacemos | Explore our research · Explorar nuestra investigación |
| `/research` | Build a better farm · Construir una granja mejor | Meet Iris · Conoce Iris |

**These are my wording, not marketing's** — happy to swap them for whatever copy
you'd prefer.

### Two defects found while moving the code

- `/research` linked to `index/introducing-iris` with **no leading slash**.
  `toSafeHref` rejects bare relative paths, so once routed through
  `MarketingPillLink` the CTA would have silently rendered as an unlinked
  `<span>`. Normalized to `/index/introducing-iris`, which is what
  `menu-items.tsx` and the `/iris` redirect already use. The new component test
  pins this behaviour so it can't regress.
- The about *vision* test matched its link with the substring regex
  `/Our research/i`, which the new closing label also satisfies. Made it exact.

### Unrelated fix included: `bun.lock`

`bun.lock` carried a **duplicate `"knip/zod"` key**, which makes the file
unparseable. Bun responds by printing `Ignoring lockfile` and resolving every
install from the `package.json` ranges instead — so dependency pinning has been
silently off, repo-wide, since that entry appeared. CI runs plain `bun i`, so it
has been resolving fresh on every run; it's green today only because current
releases happen to work. A fresh resolve on my machine picked up newer
`@hookform/resolvers`/`zod` typings and broke `type-check` on `accept-form.tsx`,
on a clean `main` with none of my changes.

The fix is deleting the repeated line. Happy to split it into its own PR if you'd
rather keep this one focused — it's in here because `bun validate` cannot run
locally without it.

### Tests

`marketing-blocks` had no tests and no stories at all; this adds the first of
each. Full `bun validate` passes (745 site + 55 db tests) and `build-storybook`
compiles.

## Checklist

### DIFF
--- a/apps/site/src/app/(unauthenticated)/[locale]/(marketing)/about/components/about-landing.test.tsx
+++ b/apps/site/src/app/(unauthenticated)/[locale]/(marketing)/about/components/about-landing.test.tsx
@@ -63,7 +63,8 @@ describe('AboutLanding', () => {
       })
     ).toBeInTheDocument();
 
-    const cta = screen.getByRole('link', { name: /Our research/i });
+    // Exact match: the closing CTA label also contains "our research".
+    const cta = screen.getByRole('link', { name: 'Our research' });
     expect(cta).toBeInTheDocument();
     expect(cta.querySelector('svg')).not.toBeNull();
   });
@@ -114,11 +115,14 @@ describe('AboutLanding', () => {
     ).toBeInTheDocument();
   });
 
-  it('renders navigation link to What We Do', () => {
+  it('renders the closing call to action pointing at research', () => {
     renderWithNextIntl(<AboutLanding />);
 
     expect(
-      screen.getByRole('link', { name: /What we do/i })
+      screen.getByRole('heading', { level: 2, name: /What we do/i })
     ).toBeInTheDocument();
+    expect(
+      screen.getByRole('link', { name: /Explore our research/i })
+    ).toHaveAttribute('href', '/research');
   });
 });
--- a/apps/site/src/app/(unauthenticated)/[locale]/(marketing)/about/components/about-landing.tsx
+++ b/apps/site/src/app/(unauthenticated)/[locale]/(marketing)/about/components/about-landing.tsx
@@ -13,6 +13,7 @@ import { HiArrowLongRight } from 'react-icons/hi2';
 import type { ReactNode } from 'react';
 import CompetenciesSection from './competencies-section/competencies-section';
 import Partners from './partners';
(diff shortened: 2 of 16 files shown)


akeight:
@Orlando275 Thanks! We have the bun.lock update in main already. It should be the same, but go ahead and pull from there just to be sure. :)

Orlando275:
@akeight Yep, It's the same. Thanks, Ally. Could you review?

akeight:
Thank you! This is perfect!
While we're in these files, we're going to change and update the text and links. Vincent or I will reach out with all those details once he finalizes that. :)

akeight:
Hey @Orlando275 !
For the text here, Vincent wants it to match the Figma mockup design.

`/research` navigates to /research/index
Screenshot 2026-09-02 at 7 29 38 PM

`/about` navigates to /careers
Screenshot 2026-09-02 at 7 30 02 PM

PR #1129, opened 2026-08-30
feat(site): honor prefers-reduced-motion in smooth scrolling

andrewdack:
## Description

Make it so that smooth scrolling is disabled when user prefers-reduced-motion.

Lenis handles scrolling through JavaScript, so the existing reduced-motion CSS did not disable its inertia. We were also still using the deprecated @studio-freight/lenis package, which no longer receives updates.

Changes

* Migrate @studio-freight/lenis → lenis@^1.3.26.
* Enable respectReducedMotion so reduced-motion users get immediate, 1:1 scrolling.
* Keep lerp: 0.07 and remove duration/easing to preserve the existing scroll feel under Lenis 1.3.
* Import the Lenis 1.3 stylesheet.
* Rename the global from window.lenis → window.__lenis to avoid a collision with Lenis 1.3.

No linked issue.

QA

bun run validate passes: type-check, lint, 755 tests, and build.

Manual QA:

* Enable System Settings → Accessibility → Display → Reduce Motion and confirm scrolling has no inertia.
* Note: my manual QA was done thru macOS system settings & firefox.
* Confirm route changes still jump to the top.
* With reduced motion disabled, confirm general scroll behavior still feels unchanged.

Checklist

* Tests updated where applicable.
* Inline docs included where applicable.
* Accessibility considered.
* Conventional commits used where appropriate.

Lenis issue referenced: https://github.com/darkroomengineering/lenis/issues/534


akeight (approved):
LGTM!

PR #1105, opened 2026-08-24
feat(site): refactor about page partners section

andrewdack:
## Description

Fixes #1103

Updates the About page partner section to match the issue reference. Preserves the existing rotating partner groups.

### What changed

- Updates the localized partner heading and description text.
- Centers the heading and description and adjusts their weight, wrapping, and spacing.
- Reorders the first partner-logo group to match the reference.
- Resizes the logos to match the reference, with tailored BDA and CCOF sizing.
- Uses a two-column, three-row mobile grid and a centered desktop row.
- Adds accessible partner names and feature-local typed logo props.
- NOTE: es.json was updated using Google Translate

### Testing

- Updates the About-page test for the new heading and accessible logo names.
- bun run validate passes formatting, types, lint, 651 tests, and builds.

## Checklist

- [x] Tests included where applicable.
- [x] Inline docs included where applicable.
- [x] Modified components are accessible.
- [x] Conventional commit format used.

PR #1149, opened 2026-09-04
fix(site): fix a11y and localization gaps in the about partners section

Kenyang1:
## Description

No linked issue — found during a codebase audit of the `/about` partners section.

Three defects in one component, plus two follow-on fixes surfaced by review:

### 1. Two partner logos were mislabeled

The alt text added in #1105 named two organizations incorrectly. Verified against the logo artwork itself and the organizations:

| Logo | Before | After |
| --- | --- | --- |
| `ofa.png` | Organic **Farming** Association | Organic **Farmers** Association (the logo spells it out) |
| `whyregenerative.png` | Why Regenerative **Agriculture** | Why Regenerative (the org's actual name, per whyregenerative.com) |

### 2. Hard-coded English on the Spanish page

"And Other" / "Notable Partners" were literal strings in the component, so `/es/about` showed English. They now come from the `about` message catalogs (`partners.tagline.0/1`), with Spanish copy ("Y otros" / "socios destacados"). The two-key split matches the two grid cells the phrase occupies; this constraint is documented at the point of use.

### 3. The rotation ignored `prefers-reduced-motion`

The 4.5s logo auto-rotation ran unconditionally. It is now gated on the house `useMediaQuery('(prefers-reduced-motion: reduce)')` hook — the same hydration-safe pattern `competencies-section` uses — rather than framer-motion's `useReducedMotion`, which returns `null` on the server but resolves synchronously on the client and would have produced a guaranteed hydration mismatch for exactly the reduced-motion users the fix targets (verified against the installed framer-motion 12.38 source).

Reduced-motion visitors get all ten logos plus the tagline rendered statically, and the matchMedia subscription stops the rotation live if the preference changes mid-session.

**Bonus from static-first:** the server HTML now contains all ten partner names and the tagline. Previously the four group-two partners never existed in the server HTML at all — invisible to crawlers and initial screen-reader passes.

### Notes for review

- `AnimatePresence` gets `initial={false}` because the rotating branch mounts after the static layout has already painted — without it, visible logos blink to transparent and fade back in on the post-hydration handoff.
- `about-landing.test.tsx` now installs the house `stubMatchMedia` helper, since `Partners` reads matchMedia through the real hook there and jsdom ships none.
- New `partners.test.tsx` (7 tests) pins all ten accessible names exactly (guarding the corrected ones), the rotation, the catalog-backed tagline (including that the es keys exist and differ from en), the SSR static shell, and that no interval is scheduled under reduced motion.
- A visible pause control (full WCAG 2.2.2 for users without the OS preference) was deliberately left out — that's a design/product call; happy to open a follow-up issue.

## QA

`bun run validate` passes: format, type-check, lint, 767 tests, build. Local note: the build/pre-push step needed a syntactically valid `DATABASE_URL` supplied to the process, because my machine's value fails the strict parsing introduced in #1092 — unrelated to this change; CI runs with its own env.

## Checklist

## Before Img
beforepartnerses

## After Img
afterpartnerses
afterpartnersenreducedmotion
afterpartnersesreducedmotion


akeight (approved):
LGTM. Thanks!

akeight:
@Kenyang1 when you resolve the merge conflicts here you can merge this :)

PR #1127, opened 2026-08-27
fix(caption): add alt caption beneath the image on /about

andrewdack:
Fixes #1120

  ## Description
  - 14px left aligned caption text added below the first image on /about
  - Added i18n translations to `en.json` and `es.json` as `vision.imageCaption`
  - Updated the second image's alt from hard-coded "Meadow" to be more descriptive -> put into
  useTranslations json as `headerImage.alt`
  - Caption is a plain `figure`/`figcaption` in `about-landing.tsx`. That image has no parallax, so no
  positioning logic is needed and the shared `header-img` component is untouched.
  - Caption colour is `text-foreground/70` (5.54:1, passes WCAG AA) rather than `text-muted-foreground`.
  That utility is currently dead — `--color-muted-foreground` is missing from the `@theme inline` block
  in `globals.css`, so Tailwind emits no rule for it despite ~27 files referencing it. Worth a separate
  issue.

  ## Checklist

  - [x] You've included unit or integration tests for your change, where applicable.
  - [x] You've included inline docs for your change, where applicable.
  - [x] Any components that you've modified are accessible.
  - [x] You've used [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/) where
  appropriate

PR #876, opened 2026-05-10
fix: update page routes and change background

yousefalwahami:
## Description

This updates two things:
- Change the whole site (internal + marketing) to white (`#ffffff`)
- Update the page routes such that `/what-we-do` -> `/research` and `/who-we-are` -> `/about`

Technical/Specific things:
- Redirects are permanent redirects from `/[locale]/who-we-are` (and `:path*`) -> `/about`, and `/[locale]/what-we-do` -> `/research` in `apps/site/next.config.ts` (SEO / bookmarks).
- Sitemap also reflects these changes.
- i18n reflects these changes.
- Updated theme in `global.css` under internal and marketing

## Checklist

- [x] You've included unit or integration tests for your change, where applicable.
- [x] You've included inline docs for your change, where applicable.
- [x] Any components that you've modified are accessible.
- [x] You've used [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/) where appropriate

PR #1151, opened 2026-09-04
fix(site): register the missing theme color tokens with Tailwind

Kenyang1:
## Description

No linked issue — this is the theme-level defect deferred by two recent PRs: #1127 ("`--color-muted-foreground` is missing from the `@theme inline` block … Tailwind emits no rule for it despite ~27 files referencing it. Worth a separate issue.") and #1128 ("the root cause … is a theme-level fix well outside this issue").

### The defect

In Tailwind v4, a color utility only exists if its token is declared in `@theme`. Our `@theme inline` block mapped 11 of the shadcn variables and silently dropped the rest — so ~120 class references across ~30 files (`text-muted-foreground` ×68, `bg-primary`, `text-destructive`, `ring-ring`, …) emitted **no CSS at all**. Every form error on the site rendered in body color instead of red; every secondary-text hierarchy the authors wrote never displayed; the shadcn default button rendered with no background.

### The fix

`globals.css` now maps **8** additional tokens: `muted-foreground`, `primary`, `primary-foreground`, `destructive`, `destructive-foreground`, `secondary-foreground`, `card-foreground`, `ring`.

**`input` is deliberately NOT mapped** (comment in the CSS): #876 redefined `--input` as a white control *fill*, but shadcn consumes `input` as the *border* token — mapping it would paint every `border-input` (outline buttons, bare inputs) white and erase them. Unmapped, those borders keep their current `currentColor` fallback, pixel-identical to today.

The existing `--color-secondary → var(--background-secondary)` and `--color-popover → var(--background)` mappings are left alone on purpose — git history shows both are deliberate overrides from the #876 "whole site to white" change.

### What activates (the intended design finally rendering)

- Validation/error text across signup, login, the apply flow, and marketing dynamic forms turns `#dc2626` red (4.5:1) — previously indistinguishable from body copy.
- All secondary text (~68 sites: field descriptions, dialog/sheet descriptions, page-hero subtitles, widget empty states, placeholders, calendar out-of-month days, inactive sidebar links) renders `#555555` on white (7.46:1).
- Default-variant Buttons, checked Checkboxes, the Switch track, and Calendar selected days gain their solid `#2a2727`/`#fcfbf8` styling (14.3:1). The most visible instance is the marketing header Sign-up CTA, which becomes a solid pill on every marketing page — flagging for design sign-off; it is `rounded-md` where the brand pill is usually `rounded-full`.
- The destructive Button variant (reminder delete confirm) turns red.

### Call-site repairs shipped in the same change

A full impact audit of every consumer found surfaces that were authored *around* the dead utilities and would break on activation. Fixed here:

| Surface | Without the fix | Fix |
| --- | --- | --- |
| Account side-menu **Log out** | `text-foreground` label on newly-solid `bg-primary` → invisible | `variant="ghost"` |
| Widget delete ✕, platform new-tab ＋ | icon-on-dark chip → invisible / off-design | `variant="ghost"` |
| Apply **Yes/No** address toggles | both solid dark; selected-underline affordance lost | `variant="ghost"` |
| Apply resend-verification icon | icon invisible on dark chip (sibling was already ghost) | `variant="ghost"` |
| Apply terms refresh icon | dark chip inside a red alert | `variant="ghost"` |
| Footer **Manage Cookies** trigger (+ latent twin in `public-inquiry-modal`) | inline text → dark chip | `variant="link"` |
| Cookie **Save** | outline pill → filled, border swallowed | `variant="outline"` |
| Cookie **Switch** | ON would flip green → near-black (`data-[state=checked]:bg-primary` outspecifies the caller's plain `bg-green-500`) | caller uses `data-[state=*]` classes; ON=green / OFF=gray exactly as before |
| Dialog close ✕ | `text-muted-foreground` @70% opacity → 2.8:1 at rest | dropped the rest-state opacity; kept a `hover:opacity-70` affordance |
| 10× `text-muted-foreground/70` (5 files) | would flatten to ~#888 (3.54:1, AA fail) — on user-typed input text in places | dropped the `/70` → 7.46:1 |

### Verification

- Impact audit: 4 agents enumerated every consumer of every activating token with computed colors and contrast ratios; the fixes above are its confirmed regression list.
- Adversarial verification: 2 agents replayed each edited call site through the installed `tailwind-merge`/`cva` (confirming, e.g., the Switch caller's classes evict the component's, and ghost hover interplay), and independently swept all 101 `<Button` call sites for missed regressions. Both pass.
- Live browser checks on the dev server: `text-muted-foreground` computes `rgb(85,85,85)`, `text-destructive` computes `rgb(220,38,38)`, footer trigger stays transparent inline text.

### Known issue (pre-existing, deliberately out of scope)

The platform tab strip (`ui/tabs.tsx` `bg-muted` + the authenticated tab bar) pairs the inverted near-black `--muted` with the newly-live `#555555` labels: ~1.9:1. Before this change those labels were `#2a2727`-on-`#2a2727` — fully invisible — so this is a pre-existing broken surface made slightly less broken, not a regression. The root cause is the `--muted` inversion #1128 documented; fixing it restyles six surfaces (skeletons, separators, tab strip, chart fill, pill hover) and is queued as a follow-up.

### Also included: `bun.lock` sync

Dependabot's 2f50af2 (#1138) bumped `@vitest/coverage-v8` to 4.1.11 in `apps/site/package.json` without regenerating the lockfile, so `bun i` on current main dirties the tree. The second commit here is the mechanical sync — happy to split it out if you'd rather keep this PR single-purpose.

## QA

format:check, type-check, lint, and build all pass locally (build fed a placeholder `DATABASE_URL`, as in #1149 — this machine's value fails the #1092 parser; CI uses its own env). The full test suite is flaky on this machine under full parallelism: across three runs, a *different* set of 4–6 tests hit the 5s timeout each time (including tests untouched by this diff like "renders without crashing"), and **every failing file passes when run in isolation** — including the edited `management-zone-form` suite (22/22). The push used the documented `--no-verify` ritual from CLAUDE.md; CI is the authoritative gate.

## Checklist

- [x] You've included unit or integration tests for your change, where applicable. (No behavior changed that jsdom can observe — the audit confirmed no test asserts on the previously-dead rendering; existing suites for every touched component pass.)

## Before Investors Img
beforeinvestors

## After Investors Img
afterinvestors

## Before Home Img
beforehome

## After Home Img
afterhome


akeight:
Can you include some screenshots of what you're updating? Some of these files look like they're legacy files or things we are currently working on refactoring, so I want to be sure without pulling them all up right now.

PR #1166, opened 2026-09-11
fix(site): restore broken approved-applicant signup links

Jose-Gael-Cruz-Lopez:
## Description

Every `/signup` approval link was rendering **"There was an error with authentication"** instead of the password form. Reported as "magic links are down"; it affected all of them, not one bad token.

Example: `/signup?application_id=4&token=8abd4716-…`

### Root cause

The server render was never at fault. Production's RSC payload carried the correct prefill for the applicant — the page died during **client hydration**.

Deminifying the frame from the live page landed on next-intl's `useIntlContext`, which does `throw Error(void 0)`. The empty message is why the error boundary showed a blank line under its heading. The caller is next-intl's `Link`:

`unauthenticated-header.tsx` → `<ToddHeader localeAware />` → locale-aware `Link` → `useLocale()` → no provider → throw

`/signup` sits outside `[locale]`, and `NextIntlClientProvider` is only mounted under `[locale]`. `ToddHeader`'s own JSDoc warns about exactly this case.

Timeline: the signup layout landed 2026-06-14 (816fd09f). `localeAware` was added to the shared header 2026-06-17 (13a747a7, "preserve locale on home navigation") — correct for its two localized callers, fatal for the one unlocalized caller.

**Fix:** `localeAware` is now a prop on `UnauthenticatedHeader`, defaulting off. `(contact)/layout.tsx` and `not-found.tsx` pass it explicitly, so 13a747a7's behaviour is preserved.

### Second bug, same journey

Found while reproducing. `signupToken` is a `uuid` column and `id` is `int4`, so Postgres **aborts the statement** (`22P02` / `22003`) rather than matching zero rows. A token truncated by an email client, or an oversized id, produced the same crash page instead of the expired-link screen.

Guarded in `form-submission-signup.ts` (covers every caller) plus the id parse in `page.tsx`. Malformed links now fall through to the gate with its resend button.

### Why CI never caught this

`vitest.setup.intl.js` globally mocks `next-intl` (`useLocale → 'en'`) and `@/i18n/config` (`Link → <a>`), and every render helper wraps in `NextIntlClientProvider`. No test in the repo *can* reach a missing-provider path. The new tests override those stubs locally.

### Verification

- Ran the fixed production build locally: all eight malformed variants render "This signup link is invalid or expired" with the resend button, zero Postgres errors, no-params still redirects to `/contact`, no error boundary.
- Each new test was confirmed to **fail** against the unfixed code, so none is vacuous.
- `type-check`, `lint`, `build` green. 759 site + 63 db tests passing.

### Follow-ups (not in this PR)

1. `components/common/button/button.tsx` renders the same locale-aware `Link` and is reachable from `(authenticated)/layout.tsx`, which per `CLAUDE.md` is also not internationalized. Same class of bug, different route tree — wants its own verification.
2. The test-setup blind spot above is worth addressing repo-wide.

## Checklist

https://claude.ai/code/session_01PrBdUGDaSL6nyQ3egQdK3D


akeight (approved):
Tested end to end. Everything works as it should. Thanks!

